PR #3, opened 2020-03-10
Implement/log shared lib

bit-bender:
Implement kmyth_log() functionality as a shared library (kmythlogger). This moves some "supporting functionality" code out of kmyth itself and allows easier reconfiguration and potential reuse of this logging utility for other applications.

The logging configuration data was placed into a struct and "setter" functions were added to the API to support modifications to default values. The default values and function names were made more generic. For example, kmyth_log() became log_event(). To avoid changes to every kmyth_log() call in the source code, a macro was used to map kmyth_log() to log_event(). The LOGINFO macro was then incorporated into the kmyth_log() macro.


kmyth_log/include/kmyth_log.h
```diff
@@ -10,152 +10,53 @@
...
- *
- * @brief   Handle error code (RM_RC (TPM2_RC_HANDLE + TPM2_RC_P + TPM2_RC_1))
- *
- * Found in tpm2-abrmd-2.0.2/src/resource-manager.c, and is returned when
- * a handle doesn't map to one that the resource manager is managing.
- * TPM2_RC_HANDLE indicates that the error is related to a handle,
- * TPM2_RC_P signifies that is is a parameter, and
- * TPM2_RC_1 specifies that it is the first parameter.
+/**
+ * @brief maximum length (in chars) of path string for the application log file
```
b-carr, line 52: Does this include the terminating null character? This same comment applies to the descriptions of the maximum lengths for app name and version.

kmyth_log/include/kmyth_log.h
```diff
@@ -194,7 +90,7 @@
  * the log file destination will not be accessible and the user will see
  * the log entries on the console.
  */
-#define KMYTH_APPLOG_OUTPUT_MODE_DEFAULT 1
+#define APPLOG_OUTPUT_MODE_DEFAULT 1
```
b-carr, line 93: Given that this is the kmyth_log library, why did you take KMYTH out of the name here?

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+};
+
+//############################################################################
+// set_app_name()
+//############################################################################
+void set_app_name(char *new_app_name)
+{
+  bool truncated = false;
+
+  if (strlen(new_app_name) <= MAX_APP_NAME_LEN)
```
b-carr, line 37: This should probably use strnlen and set the maximum value to the appropriate maximum size from the header. Since I think I'm going to make this comment a few places from here on out I'll just comment "strnlen"

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+  {
+    log_settings.app_name_len = strlen(new_app_name);
```
b-carr, line 39: strnlen

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+}
+
+//############################################################################
+// set_app_version()
+//############################################################################
+void set_app_version(char *new_app_version)
+{
+  bool truncated = false;
+
+  if (strlen(new_app_version) <= MAX_APP_VERSION_LEN)
```
b-carr, line 68: strnlen

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+  {
+    log_settings.app_version_len = strlen(new_app_version);
```
b-carr, line 70: strnlen

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+    fprintf(stderr, "truncated to \"%s\"\n", log_settings.app_version);
+  }
+}
+
+//############################################################################
+// set_applog_path()
+//############################################################################
+void set_applog_path(char *new_applog_path)
+{
+  if (strlen(new_applog_path) <= MAX_APPLOG_PATH_LEN)
```
b-carr, line 98: strnlen

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+  {
+    log_settings.applog_path_len = strlen(new_applog_path);
```
b-carr, line 100: strnlen

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+    // do nothing if invalid, but warn user
+    fprintf(stderr, "set_applog_severity_threshold(): ");
+    fprintf(stderr, "input (%d) invalid - unchanged ", new_severity_threshold);
+    fprintf(stderr, "(%d)\n", log_settings.applog_severity_threshold);
+  }
+}
+
+//############################################################################
+// set_syslog_facility()
+//   - valid values: (0<<3) - ((LOG_NFACILITIES-1)<<3)
```
b-carr, line 157: Why is 0 given here as (0<<3)?

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+//############################################################################
+void set_syslog_facility(int new_syslog_facility)
+{
+  if ((LOG_FAC(new_syslog_facility) >= 0) &&
+      (LOG_FAC(new_syslog_facility) <= (LOG_NFACILITIES - 1)))
```
b-carr, line 162: This doesn't appear to be consistent with the comment, which suggests the (LOG_NFACILITIES-1) should be left shifted 3.

kmyth_log/src/kmyth_log.c
```diff
@@ -0,0 +1,350 @@
...
+    asprintf(severity_str_out, "NOTICE");
+    break;
+  case LOG_INFO:
+    asprintf(severity_str_out, "INFO");
+    break;
+  case LOG_DEBUG:
+    asprintf(severity_str_out, "DEBUG");
+    break;
+  default:
+    asprintf(severity_str_out, "CUSTOM(%d)", severity_val_in);
```
b-carr, line 227: How can a custom level be defined? Looking at log_even I guess using negative severity_val_in?

tpm2.0/src/getkey/main.c
```diff
@@ -8,10 +8,10 @@
  */
 
 #include "tls_util.h"
+#include "tpm2_kmyth_global.h"
 #include "tpm2_kmyth_io.h"
 #include "tpm2_kmyth_misc.h"
 #include "tpm2_kmyth_seal.h"
-#include "kmyth_log.h"
```
b-carr, line 14: I think we've gone both ways on this one, but should we leave the log include here because we're calling functions from that API?

kmyth_log/include/kmyth_log.h
bit-bender: No, the maximum length parameters (MAX_APP_NAME_LEN, MAX_APP_VERSION_LEN, MAX_APPLOG_PATH_LEN, and MAX_LOG_MSG_LEN) do not include the terminating null character.

I think I verified that everywhere I use this value to specify array sizes for the corresponding string values, I add one to the maximum length value to accommodate for the string terminator.

When formatting the log message string (in log_event()), I pass MAX_LOG_MSG_LEN+1 as the count parameter to vnsprintf(), because the documentation indicates that this should include the terminating null character.

When using strnlen(), I believe that I correctly expect the returned length to not include the terminating null character.

kmyth_log/include/kmyth_log.h
bit-bender: I think that, as part of putting the logging routines into a separate library, I wanted to make it more generic. As we kept kmyth in the name of the library, though, this was probably an inconsistent goal.

Therefore, I have changed the following variables back to include KMYTH in the name:
    KMYTH_APPLOG_OUTPUT_MODE_DEFAULT
    KMYTH_APPLOG_SEVERITY_THRESHOLD_DEFAULT

I also changed DEFAULT_APP_NAME to "kmyth" and DEFAULT_APPLOG_PATH to "/var/log/<DEFAULT_APP_NAME>.log" (this used to be default_app and /var/log/default_app.log).

kmyth_log/src/kmyth_log.c
bit-bender: Yes, I think I corrected these. I also put the value of the strnlen() call into a temp_len variable rather than call strnlen multiple times.

kmyth_log/src/kmyth_log.c
bit-bender: see above

kmyth_log/src/kmyth_log.c
bit-bender: see above

kmyth_log/src/kmyth_log.c
bit-bender: see above

kmyth_log/src/kmyth_log.c
bit-bender: see above

kmyth_log/src/kmyth_log.c
bit-bender: see above.

kmyth_log/src/kmyth_log.c
bit-bender: So, in syslog, the priority and facility values are combined. The lowest three bits are the priorty value (0-7) and the remaining upper bits are the facility. Although there are many more bits, typically the number of facility values (LOG_NFACILITIES) is 24. Since I pass this value to LOG_FAC(), which right-shifts it by three bits, before evaluating its value, the valid, pre-decoded facility values are (0 << 3) to (23 << 3). I tried to make this clearer in the comments.

kmyth_log/src/kmyth_log.c
bit-bender: Because LOG_FAC() right shifts new_syslog_facility by three bits, I think this is consistent. If the comments do not make this clearer or if you still believe this is wrong, though, let me know and I will look at it again.

kmyth_log/src/kmyth_log.c
bit-bender: A "custom" level greater than 7 would be inconsistent with syslog defined levels and would never satisfy the severity threshold. I guess a negative severity_val_in would meet the threshold, but would the inconsistent with the syslog defined levels and I don't know what value they might add.

In the log_event() function, therefore, I added a line that passes the input severity parameter to LOG_PRI() before using it. This will mask out all but the lowest three bits, forcing the value to the range 0-7. The default case in this switch statement, therefore, should actually never occur. Just in case, however, I changed the severity string result to "INVALID SEVERITY (%d)".

tpm2.0/src/getkey/main.c
bit-bender: I added #include <kmyth_log.h> to all of the main.c files, because we use the API to configure the logging (e.g., app name, log-file path, ...). In include/tpm2_kmyth_global.h, we map the kmyth_log() macro to the log_event() function in the API, so I it needs to remain there. I don't think any of the other files use the API directly. I think they just call the kmyth_log() macro.

bit-bender:
In testing, I am getting failures when the TPM is trying to generate a new SRK: kmyth-0.0.0 ERROR - src/util/tpm2_kmyth_object.c(tpm2_kmyth_create_object:348) Tss2_Sys_CreatePrimary(): rc = 0x00090006, mu:A buffer isn't large enough ... exiting. I think this is an issue with my test environment, rather than kmyth, but I am trying to work through it. The logging (using the shared library version) up to the point of failure, however, seems to be working OK.

b-carr:
I've seen this exact issue intermittently.

PR #10, opened 2020-08-11
Update the TPM2.0 README with dependency install instructions

PeterHamilton:
This change adds a series of install instructions for each of the dependencies needed to test and run Kmyth, including the tpm2-tss, tpm2-tools, and tpm2-abrmd libraries, along with the IBM TPM 2.0 Emulator. The instructions are intended for use on CentOS 8.

### DIFF
--- a/tpm2.0/README.md
+++ b/tpm2.0/README.md
@@ -78,6 +78,163 @@ Table of Contents
 
 ### Building
 
+#### Building the Dependencies
+
+First, install as many of the above listed dependencies as you can.
+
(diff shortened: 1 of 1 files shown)


PeterHamilton:
It may make more sense to move these additions to a new `INSTALL.md` file, given that section is getting long in the current `README.md`. Just let me know if that's preferred.

b-carr:
I think moving the new additions to a new `INSTALL.md` makes the most sense. The README is getting pretty long.

PR #23, opened 2020-08-27
Removed support for TPM 1.2

PircDef:
Code for TPM 1.2 was removed  
The tpm2.0 directory was renamed simply tpm2


PircDef:
Good call, it's removed.

Issue #4, opened 2020-03-11
Restructure build using recursive makefiles


bit-bender:
As a step to make the code and build process more modular and portable, it will probably help to revamp the current single, top-level makefile approach by instead applying a hierarchical (recursive) makefile schema.

b-carr:
As part of this we should add build flags to turn on/off optional features. The only one I have immediately in mind is the KMIP `getkey`. We should be able to build against any (or all) available `getkey` backends but if the user doesn't want a particular backend we shouldn't require it be present just to build others.

bit-bender:
Although we went a different direction (e.g., a centralized Makefile), pull request #24 reworked and improved the build process. Therefore, I am closing this general "make the build process more modular and portable" issue with the expectation that any further necessary refinements to the build process will be covered by new and more specific issues.

PR #24, opened 2020-08-28
Improvement/centralize makefile

PircDef:
* Updated Makefile to build the logger then the kmyth tool
* Removed the logger-specific Makefile
* Updated doc to reflect CentOS 8
* Removed the OpenSSL dir options needed for CentOS 7


PircDef:
This should address what we wanted with issue #4

tpm2/Makefile
```diff
@@ -1,13 +1,10 @@
 # Makefile for KMYTH using TPM 2.0
 
-# Choose OpenSSL 'base' path (header file and library path prefix)
-OPENSSL_DIR = $(MY_OPENSSL_DIR)#         custom shell environment variable
-
 # Specify the 'libkmythlogger' library path
-KMYTH_LOG_DIR = $(MY_KMYTH_LOG_DIR)#       custom shell environment variable
+LOGGER_DIR ?= logger#       		       Set a default path to the logger
```
bit-bender, line 4: This path to the logger should be absolute. With this relative path, I believe that if the applications are invoked from any directory other than <path>/kmyth/tpm2, the logger library will not be found. One way to potentially fix this is by using "LOGGER_DIR ?= $(abspath logger)" instead.

tpm2/Makefile
```diff
@@ -73,16 +80,33 @@ UTIL_SOURCES = $(wildcard $(UTIL_SRC_DIR)/*c)
...
 OBJECT_DIRS = $(SEAL_OBJ_DIR)
 OBJECT_DIRS += $(UNSEAL_OBJ_DIR)
 OBJECT_DIRS += $(GETKEY_OBJ_DIR)
 OBJECT_DIRS += $(UTIL_OBJ_DIR)
+OBJECT_DIRS += $(LOGGER_OBJ_DIR)
+DOC_DIRS = $(DOC_DIR)
+DOC_DIRS += $(LOGGER_DOC_DIR)
 
-all: pre kmyth-seal kmyth-unseal kmyth-getkey
+all: pre logger kmyth-seal kmyth-unseal kmyth-getkey
```
bit-bender, line 106: Since 'all' is not a file-based target, I think it should be phony (.PHONY: all).

tpm2/Makefile
```diff
@@ -100,25 +124,30 @@ pre:
 	rm -f include/*.h~
 	mkdir -p $(BIN_DIR)
 	mkdir -p $(OBJECT_DIRS)
-	mkdir -p $(DOC_DIR)
+	mkdir -p $(DOC_DIRS)
+	mkdir -p $(LOGGER_LIB_DIR)
 
-doc: $(HEADER_FILES) $(UTIL_SOURCES) $(TOOL_SOURCES) $(SEAL_SOURCES) $(UNSEAL_SOURCES) $(GETKEY_SOURCES) Doxyfile
+doc: $(HEADER_FILES) $(UTIL_SOURCES) $(TOOL_SOURCES) $(SEAL_SOURCES) $(UNSEAL_SOURCES) $(GETKEY_SOURCES) $(LOGGER_HEADERS) $(LOGGER_SOURCES) Doxyfile
```
bit-bender, line 130: Since 'doc' is not a file-based target, it should be phony (.PHONY: doc). If it is not, when the Makefile sees that the doc directory exists, it reports that it is up to date, even though no documentation has been generated or there is a need to generated changed documentation.

tpm2/INSTALL.md
```diff
@@ -47,7 +47,7 @@ these dependencies can likely be met using your package
 manager. [libkmip](https://github.com/OpenKMIP/libkmip) must be
 installed from source.
 
-##### CentOS 7 (Red Hat 7) Commands
+##### CentOS 8 (Red Hat 8) Commands
 
 ```yum install openssl openssl-devel glibc gcc libffi-devel indent```
```
b-carr, line 52: For CentOS 8 `indent` got moved to the EPEL repo.

tpm2/Makefile
b-carr: Would it be better to bundle the logger into the shared library files rather than trying to deal with making sure it can find `libkmythlogger` at runtime?

tpm2/Makefile
```diff
@@ -100,25 +124,30 @@ pre:
...
 	doxygen Doxyfile
 
+$(LOGGER_OBJECTS): $(LOGGER_SOURCES) $(LOGER_HEADERS)
+	$(CC) $(LOGGER_CFLAGS) -I$(LOGGER_INC_DIR) $< -o $@
+
 $(SEAL_OBJ_DIR)/%.o: $(SEAL_SRC_DIR)/%.c | $(SEAL_OBJ_DIR)
```
b-carr, line 136: Do we want to be able to build the library objects without having to build the executables?

tpm2/Makefile
PircDef: I added .PHONY to all, pre, and doc

tpm2/INSTALL.md
b-carr: The Makefile still uses indent. We should either remove it from there (and so also from the Ubuntu installation instructions) or mention you need it but it's in EPEL so you have to activate that first.

b-carr (approved):
Since the focus is on the Makefile I'm approving this (especially since it makes installation easier.) But I'm also opening an issue to make sure the README and INSTALL files are updated appropriately.

bit-bender (approved):

PR #26, opened 2020-09-08
Sealing on empty file patch

wmillstone:
Added a check in `kmyth_wrap_input` in `src/util/tpm2_kmyth_seal.c` to validate that the file being sealed is not empty.

### DIFF
--- a/tpm2/src/util/tpm2_kmyth_seal.c
+++ b/tpm2/src/util/tpm2_kmyth_seal.c
@@ -667,7 +667,7 @@ int kmyth_wrap_input(char *inPath,
   kmyth_log(LOG_DEBUG, "read in %d bytes of data to be wrapped", data_length);
   
   // validate non-empty plaintext buffer specified
-  if (data_length == 0)
+  if (data_length == 0 || data == NULL)
   {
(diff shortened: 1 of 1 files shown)


tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -665,6 +665,14 @@ int kmyth_wrap_input(char *inPath,
     return 1;
   }
   kmyth_log(LOG_DEBUG, "read in %d bytes of data to be wrapped", data_length);
+  
+  // validate non-empty plaintext buffer specified
+  if (data_length == 0)
```
b-carr, line 670: For completeness we should also check that `data` is not `NULL`.

b-carr: I apologize, my earlier comment wasn't clear. I meant that we should check that `data` is not `NULL` in addition to checking that `data_length` is not zero before using it, I wasn't suggesting to check that `data `is not `NULL` before `free`ing it.

b-carr (commented):
@PircDef Do we want to squash these commits before merging them in?

PR #30, opened 2020-09-16
Combined logger and utils into a single libkmyth.so

PircDef:
Moved utils and logger into one .so for kmyth
Fixed the install to install only the desired headers
Removed a floating white space


tpm2/Makefile
```diff
@@ -23,36 +23,36 @@ LLIBS += -ltss2-rc#                      TPM 2.0 Return Code Utilities
 LLIBS += -lssl#                          OpenSSL
 LLIBS += -lcrypto#                       libcrypto
 LLIBS += -lkmip#                         libkmip
-LLIBS += -lkmythlogger#                  kmyth logging utility shared library
+LLIBS += -lkmyth#                        kmyth shared library
```
bit-bender, line 26: This is kind of a trivial comment, but I was looking at the GNU make manual for something else and noticed that the convention is to use LDLIBS and LDFLAGS (as opposed to LLIBS and LFLAGS). LFLAGS is actually the variable used to pass parameters to Lex.

tpm2/Makefile
```diff
@@ -23,36 +23,36 @@ LLIBS += -ltss2-rc#                      TPM 2.0 Return Code Utilities
...
 LOGGER_CFLAGS = $(CFLAGS)
-LOGGER_CFLAGS += -fPIC#                     Generate position independent code
 
 SOFLAGS = -shared#                          compile/link shared library
 
 # Specify linker flags
-LFLAGS += -L${LOGGER_DIR}/lib#          libkmythlogger
-LFLAGS += -Wl,-rpath=${LOGGER_DIR}/lib# runtime path for libkmythlogger.so
+LFLAGS += -Llib#                            libkmyth
+LFLAGS += -Wl,-rpath=lib#                   runtime path for libkmyth.so
```
bit-bender, line 45: Again, I'm not sure that this really adds anything, but, if we moved the directory variables up, we could use -L$(LIBDIR) and -rpath=$(LIBDIR) to keep it consistent if that is ever changed for some reason.

PircDef: I think if someone is modifying this for a local build and test, it should be an intentional edit to that part of the Makefile.

bit-bender: I agree that hard-coding the local build and test configuration is OK. I was just thinking that since we made a variable to parameterize other local build environment actions, we might use it here as well. I'm OK, though, with leaving it as is.

bit-bender (approved):

Issue #25, opened 2020-09-08
Clean up error handling and memory management


b-carr:
There are quite a few places in the code where memory isn't freed on failure, or where failure conditions aren't properly handled. We need to go through the entire codebase ensuring all errors are handling appropriately and all memory is freed appropriately.

wmillstone:
It looks like even when memory is freed, the freed pointer is not usually set to NULL. While not necessary, doing after a call to free() would prevent errors if the code ever attempts to free the same memory more than once.

PeterHamilton:
> It looks like even when memory is freed, the freed pointer is not usually set to NULL. While not necessary, doing this after a call to free() would prevent errors if the code ever attempts to free the same memory more than once. In most cases, this is a non-issue, but there are a few function calls that do some memory management in the background that isn't always obvious (for example `tpm2_kmyth_unseal()` in `tpm2_kmyth_seal.c`) and this would protect us against unseen errors.

Yep, I have more updates that explicitly set freed pointers to NULL for just this reason. Hoping I can get those out soon.

PR #32, opened 2020-09-18
Add error checks for getting TPM properties when freeing resources

PeterHamilton:
This change adds error handling cases to tpm2_free_resources when calling tpm2_get_properties, specifically checking if property loading fails thereby preventing session flushing. Additional error logs are added to clearly explain the error when it occurs.

Partially addresses #25


bit-bender (approved):
This corrects the functionality of the code flushing any existing TPM HMAC or policy session handles.

PR #37, opened 2020-09-28
Improvement/create generic kmyth seal unseal functions

PircDef:
Created a kmyth-seal and kmyth-unseal interface that did not require files.


tpm2/include/tpm2_kmyth_io.h
```diff
@@ -72,6 +72,137 @@ int verifyOutputFilePath(char *path);
...
+ *                                     the 'ENC DATA' block of the .ski file.
+ *
+ * @param[in]  encrypted_data_size     Size, in bytes, of the encrypted data
+ *                                     to be written to the .ski file.
+ *
+ * @return 0 if success, 1 if error
+ */
+int tpm2_kmyth_create_ski_string(uint8_t ** output,
+                                 size_t *output_length,
+                                 char *filename,
```
b-carr, line 140: Why are we tracking filename length here?

tpm2/include/tpm2_kmyth_seal.h
```diff
@@ -12,6 +12,72 @@
 #include <stdint.h>
 #include <tss2/tss2_sys.h>
 
+/**
```
b-carr, line 15: I think this functionality should be exposed in the `kmyth.h` header.

tpm2/include/tpm2_kmyth_seal.h
```diff
@@ -12,6 +12,72 @@
...
+ *                               a string)
+ *
+ * @param[out] output_data       Decrypted result (pointer to a byte buffer)
+ *
+ * @param[out] output_size       Size (in bytes) of decrypted result
+ *                               (passed as pointer to size value)
+ *
+ * @return 0 on success, 1 on error
+ */
+int tpm2_kmyth_unseal_file(char *input_path,
```
b-carr, line 75: I think this functionality should be exposed in the `kmyth.h` header.

tpm2/src/util/tpm2_kmyth_io.c
```diff
@@ -169,6 +169,523 @@ int read_arbitrary_file(char *input_path,
   return 0;
 }
 
+//############################################################################
+// tpm2_kmyth_create_ski_string
+//############################################################################
+int tpm2_kmyth_create_ski_string(uint8_t ** output,
```
b-carr, line 175: Looking at this function, the idea of having a SKI structure that can be written to file without an intermediate buffer seems really attractive.

tpm2/src/util/tpm2_kmyth_io.c
```diff
@@ -1101,3 +1618,36 @@ int print_to_stdout(unsigned char *data, size_t data_size)
...
+  if (input == NULL || input_length == 0) //nothing to concat
+  {
+    return (0);
+  }
+
+  uint8_t *new_dest = NULL;
+  size_t new_dest_len = *dest_length + input_length;
+  size_t offset = *dest_length;
+
+  if (new_dest_len < *dest_length)  //if we have an overflow
```
b-carr, line 1634: I'd prefer to test for overflow by looking at `SIZE_MAX - input_length` and comparing to `*dest_length` rather than letting the overflow happen and looking at the result.

tpm2/src/util/tpm2_kmyth_io.c
```diff
@@ -1101,3 +1618,36 @@ int print_to_stdout(unsigned char *data, size_t data_size)
...
+    return (1);
+  }
+
+  if ((new_dest = realloc(*dest, new_dest_len)) == 0)
+  {
+    kmyth_log(LOG_ERR, "Ran out of memory ... exiting");
+    return (1);
+  }
+
+  for (size_t i = 0; i < input_length; i++)
```
b-carr, line 1646: It would be cleaner to do this with a memcpy rather than iterating.

tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -30,12 +30,443 @@ extern const cipher_t cipher_list[];
 //############################################################################
 // tpm2_kmyth_seal()
 //############################################################################
-int tpm2_kmyth_seal(char *input_path,
-                    char *output_path,
+int tpm2_kmyth_seal(uint8_t * input, size_t input_length,
+                    uint8_t ** output, size_t *output_length,
                     char *auth_string,
                     char *pcrs_string,
```
b-carr, line 36: Not for this pull request, but what sort of things can go in `auth_string`? Is there any reason it should be able to contain a `NULL` byte?

tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -30,12 +30,443 @@ extern const cipher_t cipher_list[];
...
+
+  // Wrap input data -
+  //   - The data to be encrypted is contained in a file and the path to that
+  //     file is specified by the user.
+  //   - The encryption uses the symmetric 'cipher' specified by the user.
+  //   - The symmetric wrapping key used for encryption
+  //     is created as part of the call to kmyth_wrap_input().
+  kmyth_log(LOG_DEBUG, "wrapping input data");
+  size_t wrapped_data_size = 0;
+  unsigned char *wrapped_data = NULL;
```
b-carr, line 229: Why is this `unsigned char*` where other places we've used `uint8_t`?

tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -30,12 +30,443 @@ extern const cipher_t cipher_list[];
...
+  size_t wrapKey_size = get_key_len_from_cipher(cipher) / 8;
```
b-carr, line 230: Do we have any restrictions on cipher key lengths? Here we seem to assume they're some number of full octets, but what happens if somebody tries to use a cipher whose key length in bits isn't a multiple of 8?

tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -30,12 +30,443 @@ extern const cipher_t cipher_list[];
...
+
+  return 0;
+}
+
+//############################################################################
+// tpm2_kmyth_unseal()
+//############################################################################
+int tpm2_kmyth_unseal(uint8_t * input, size_t input_len,
+                      uint8_t ** output, size_t *output_len,
+                      char *auth_string, char *owner_auth_passwd)
```
b-carr, line 314: I'd really like to shift passwords to `uint8_t*`s with tracked lengths so it's possible to use key material as a password.

b-carr (approved):
I don't see any issues serious enough to justify not merging, just some places where it could be made better but those improvements can wait.

tpm2/include/tpm2_kmyth_io.h
PircDef: We don't necessarily know where the filename originates in this function, so we have no guarantee that it ends with a null terminator. We use it in the function in this call:

    concat(&out, &out_length, (uint8_t *) filename, filename_length);

tpm2/src/util/tpm2_kmyth_io.c
PircDef: Agreed and done.

tpm2/src/util/tpm2_kmyth_seal.c
PircDef: Because the underlying cipher functions use char*

PR #39, opened 2020-10-06
Improvement/create ski struct

PircDef:
Added the ski struct and centralized the code
Incorporated the new seal/unseal libraries into the mains
Added file seal/unseal to kmyth.h


tpm2/src/util/ski.c
```diff
@@ -0,0 +1,560 @@
...
+
+#include <stdio.h>
+#include <stdint.h>
+#include <string.h>
+#include <openssl/rand.h>
+
+//############################################################################
+// tpm2_kmyth_parse_ski_bytes
+//############################################################################
+int tpm2_kmyth_parse_ski_bytes(uint8_t * input, size_t input_length,
```
b-carr, line 29: What would you think about making this return the `Ski` structure? It seems weird to me that it gets passed in, but also `free`'d on error within this function.

b-carr, line 29: It really doesn't thrill me that we're modifying `input` during function calls within this function. I get why it's happening, but I think it would be clearer if we used a separate `position` pointer that gets modified.

tpm2/src/util/ski.c
```diff
@@ -0,0 +1,560 @@
...
+    return 1;
+  }
+
+  free(pcr_select_data);
+  free(sk_pub_data);
+  free(sk_priv_data);
+  free(wk_pub_data);
+  free(wk_priv_data);
+
+  //At this point the data is all formatted, it's time to create the string
```
b-carr, line 423: Since we already have the `Ski` structure in hand (and the encoded data) why aren't you allocating all the necessary memory up front?

tpm2/src/util/ski.c
```diff
@@ -0,0 +1,560 @@
...
+
+  *output = out;
+  *output_length = out_length;
+
+  return 0;
+}
+
+void free_ski(Ski * ski)
+{
+  free(ski->enc_data);
```
b-carr, line 475: Dow we not have to `free` the TPM structures in the `Ski`?

tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -462,424 +415,78 @@ int tpm2_kmyth_unseal(uint8_t * input, size_t input_len,
...
-  size_t wrapKey_size = get_key_len_from_cipher(cipher) / 8;
-  unsigned char *wrapKey = calloc(wrapKey_size, sizeof(unsigned char));
+  uint8_t *data = NULL;
+  size_t data_length = 0;
 
-  if (kmyth_wrap_input(input_path,
-                       cipher,
-                       &wrapped_data,
-                       &wrapped_data_size, &wrapKey, &wrapKey_size))
+  kmyth_log(LOG_DEBUG, "reading input file ...");
```
b-carr, line 434: Should we log the file path here?

b-carr (commented):
No major comments, I just have a few things I want to make sure we're doing the best way we can.

tpm2/src/util/ski.c
bit-bender: If we made the Ski structure the return value, however, we would have to report errors to the caller differently. Even if we pass the Ski structure as a parameter, though, I think it might be clearer to have the caller free the struct upon the function returning from an error.

bit-bender: I think the use of an offset parameter (rather than modifying the input) is a good suggestion.

bit-bender (commented):
I think this is a very good improvement and makes things cleaner.

tpm2/include/ski.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+void free_ski(Ski * ski);
+
+/**
+ * @brief Creates an 'empty' ski struct and initializes internal
+ *        sizes to 0
+ *
+ * @return A new, 'empty' ski struct
+ */
+Ski get_default_ski(void);
```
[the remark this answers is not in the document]
PircDef: I would prefer to wait to do this just now, but I'm happy to revisit if we end up with a compelling reason to do so.

tpm2/src/util/ski.c
PircDef: I have updated this to use a new Ski struct within the function and only modify 'output' when the function executes successfully.

tpm2/src/util/ski.c
PircDef: We could, but I think concat is pretty easy to read and more flexible as implemented.

tpm2/src/util/ski.c
PircDef: I don't think there are any other internal pointers needing to be freed.

bit-bender (approved):

tpm2/include/ski.h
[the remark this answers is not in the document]
b-carr: I have no idea how I managed to make this comment twice.

b-carr (approved):

PR #44, opened 2020-10-08
Add error checking to various Kmyth object utility functions

PeterHamilton:
This changes adds a series of error checking updates and fixes to the Kmyth object utility function suite. It also updates the function signature of two utilities to support returning error codes. The usage of these two functions in other files have been updated to reflect the new function signatures.

Partially addresses #25

### DIFF
--- a/tpm2/include/tpm/object_tools.h
+++ b/tpm2/include/tpm/object_tools.h
@@ -43,12 +43,12 @@
  *                             creation of a Kmyth object - passed as a
  *                             pointer to this struct
  *
- * @return None 
+ * @return 0 if success, 1 if error.
  */
-void tpm2_init_kmyth_object_sensitive(TPM2B_AUTH object_auth,
-                                      uint8_t * object_data,
-                                      size_t object_dataSize,
-                                      TPM2B_SENSITIVE_CREATE * sensitiveArea);
+int tpm2_init_kmyth_object_sensitive(TPM2B_AUTH object_auth,
+                                     uint8_t * object_data,
+                                     size_t object_dataSize,
+                                     TPM2B_SENSITIVE_CREATE * sensitiveArea);
 
 /**
  * @brief Fill in public template used to create Kmyth object.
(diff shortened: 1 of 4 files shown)


tpm2/src/util/tpm2_kmyth_object.c
```diff
@@ -34,6 +40,9 @@ void tpm2_init_kmyth_object_sensitive(TPM2B_AUTH object_auth,
   // For data, the data buffer size cannot be zero - we must populate the
   // buffer with data to be sealed and set the size to its length in bytes.
   sensitiveArea->sensitive.data.size = object_dataSize;
+  // TODO: Calling memcpy with either source or dest addresses set to NULL
+  // is technically undefined behavior. However, object_data is expected
+  // to be NULL in certain conditions.
```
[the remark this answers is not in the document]
PeterHamilton: I think this makes sense overall. I'll add the suggested change to the batch.

tpm2/src/util/tpm2_kmyth_object.c
```diff
@@ -34,6 +40,17 @@ void tpm2_init_kmyth_object_sensitive(TPM2B_AUTH object_auth,
...
+if ( (object_dataSize == 0) || (object_data == NULL) )
+{
+   sensitiveArea->sensitive.data.size = 0;
+}
+else
+{
+    sensitiveArea->sensitive.data.size = object_dataSize;
+    memcpy(&sensitiveArea.sensitive.data.buffer, object_data, sensitiveArea.sensitive.data.size);
+}
   memcpy(&sensitiveArea->sensitive.data.buffer, object_data,
```
bit-bender, line 54: Need to git rid of the memcpy() outside of the if/else. Also, I think the indent on the added code needs to be fixed and we should probably remove the TODO comment if we believe the issue is addressed by the change.

tpm2/src/util/tpm2_kmyth_object.c
```diff
@@ -34,6 +40,17 @@ void tpm2_init_kmyth_object_sensitive(TPM2B_AUTH object_auth,
...
          sensitiveArea->sensitive.data.size);
   if (object_dataSize > 0)
```
bit-bender, line 56: This debug log statement will happen if object_dataSize > 0 but object_data == NULL. If we want to keep it, we should probably either just move the kmyth_log() statement into the else block above or use sensitiveArea->sensitive.data.buffer instead of object_dataSize in the if condition.

PeterHamilton: Yep, moved it into the else block.

tpm2/src/util/tpm2_kmyth_object.c
PeterHamilton: Yep, removed the extra memcpy, the TODO comment, and fixed `sensitiveArea` field references.

PeterHamilton:
All of the above comments have been addressed. However, I had to force push the branch after rebasing on the latest additions to `main`, so the old commit hashes changed.

bit-bender (approved):

PR #45, opened 2020-10-09
Updated seal/unseal parameters to require auth sizes

PircDef:
Seal/unseal now inputs auth values as uint8_t* and corresponding sizes  
Updated underlying code to use the input sizes  
Removed several NULL checks going into kmyth_clear because kmyth_clear returns immediately when NULL is input

### DIFF
--- a/tpm2/include/kmyth.h
+++ b/tpm2/include/kmyth.h
@@ -22,30 +22,36 @@
  * @param[in]  output_path       Path to .ski file where the kmyth-seal output
  *                               will be written
  *
- * @param[in]  auth_string       Authorization string to be applied to the
+ * @param[in]  auth_bytes        Authorization string to be applied to the
  *                               Kmyth TPM objects (i.e, storage key and sealed
  *                               wrapping key) created by kmyth-seal
+ *
+ * @param[in]  auth_bytes_len    length of auth_string
+ *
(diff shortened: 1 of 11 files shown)


tpm2/include/kmyth.h
```diff
@@ -22,30 +22,36 @@
  * @param[in]  output_path       Path to .ski file where the kmyth-seal output
  *                               will be written
  *
- * @param[in]  auth_string       Authorization string to be applied to the
+ * @param[in]  auth_bytes        Authorization string to be applied to the
```
bit-bender, line 25: I will just make this comment/question once, but it applies almost universally to the header file comments for auth_bytes. In our mains, we accept an 'authString' (char *) command line input from the user, but, everywhere else, I believe that this value is treated as a more generic byte array. I'm thinking, therefore, that in the header file comments describing this parameter, we should not be using the 'authorization string' description, but rather wording to convey the potential use of a more generic 'key' or 'secret' like 'authorization bytes" or "authorization secret", or "authorization parameter" or "authorization byte array" or ???. The TPM uses the term "authorization value", so I would not overload that. If you think that "string" is a sufficiently generic description, I could probably be convinced of that as well, though.

bit-bender (approved):

PR #46, opened 2020-10-14
Added error checking for authVal creation

stacktracingx86:


tpm2/src/util/tpm2_kmyth_seal.c
```diff
@@ -326,7 +326,10 @@ int tpm2_kmyth_unseal(uint8_t * input, size_t input_len,
   if (auth_string != NULL)
   {
     kmyth_clear(auth_string, strlen(auth_string));
+    kmyth_log(LOG_ERR, "error creating authorization value (authVal) ... exiting");
```
PircDef, line 329: The function tpm2_kmyth_create_authVal() is using the contents of auth_string to create and populate objAuthValue.

The check if(auth_string != NULL) only exists to see if it needs to be freed. In a current pull request #45, we have moved away from auth_string and to auth_bytes (char* to uint8_t*), and the memory handling (the free) is the responsibility of the function calling tpm2_kmyth_unseal

stacktracingx86: I see! Thank you for reviewing, I must have been thinking of something else.

stacktracingx86: Would this still be valid after the change from auth_string to auth_bytes, after a few minor changes in it of course?

stacktracingx86: I see! Thank you for reviewing, I must've skimmed past your changes.

PircDef: Not a problem! If you come across anything else, we'll be happy to look at it.

PR #50, opened 2020-10-17
Starting Point for Unit Test Approach

bit-bender:
This is intended to be a simple and small set of unit tests. The attempt was to make it so that the structure of the CUnit test runner, how tests are organized in the test directory, the Makefile integration, etc. would be the primary focus of this pull request. To both build and run the tests, use 'make test' after the code has been built. This includes three test functions across two test suites.

### DIFF
--- a/tpm2/INSTALL.md
+++ b/tpm2/INSTALL.md
@@ -250,18 +250,16 @@ Once the dependencies are installed:
 
 1. Download the code
 
-2. Build and install the kmyth logger
-  * In the `kmyth_log` directory run *make* and *make install*
-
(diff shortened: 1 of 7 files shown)


PircDef:
The tpm2 INSTALL.md needs to have a section for the tests for CUnit and CUnit-devel

bit-bender:
I think I updated the INSTALL.md. I apologize for the number of commits, but I kept finding things in tpm2/INSTALL.md when I reviewed my changes. The force push was to correct my first commit message.

PircDef (approved):

PR #51, opened 2020-10-19
Add a CLI parameter for selecting the key server backend

### LINKED ISSUE #29: Update getkey to support both KMIP and 'simple' TLS server examples
The original version of kmyth-getkey demonstrated key retrieval from a generic TLS server. When a version of kmyth-getkey that incorporated libkmip library functionality to demonstrate a Kmyth client that retrieves keys from a KMIP server was developed, it replaced the original kmyth-getkey version.

As we would like to retain the functionality of both versions, this issue more specifically proposes modification of kmyth-getkey to support both (either of) "get_key_from_kmip_server()" and "get_key_from_tls_server()" options. Further, selection of which option to use should be via a required command line parameter.

PeterHamilton:
This change adds a new '-t|--type' CLI parameter to getkey that allows the user to select the key server backend that stores the key to retrieve. Valid options include 'kmip' and 'simple'.

Closes #29

### DIFF
--- a/tpm2/src/getkey/main.c
+++ b/tpm2/src/getkey/main.c
@@ -32,7 +32,7 @@ static void usage(const char *prog)
           "  -l or --client        Path to file containing the client's certificate.\n\n"
           "Server Information --\n"
           "  -t or --type          The type of the key server. Valid values include 'kmip' and 'simple'.\n"
-          "                        Defaults to 'simple'."
+          "                        Defaults to 'simple'.\n"
           "  -s or --server        Path to file containing the certificate\n"
           "                        for the CA that issued the server cert.\n"
           "  -c or --conn_addr     The ip_address:port for the TLS connection.\n"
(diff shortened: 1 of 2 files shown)


tpm2/src/getkey/main.c
```diff
@@ -31,6 +31,8 @@ static void usage(const char *prog)
           "  -i or --input         Path to file containing the kmyth-sealed client's certificate private key.\n"
           "  -l or --client        Path to file containing the client's certificate.\n\n"
           "Server Information --\n"
+          "  -t or --type          The type of the key server. Valid values include 'kmip' and 'simple'.\n"
```
bit-bender, line 34: The README.md file in the tpm2 directory contains usage information. We should also update that as part of this pull request to keep it current.

PeterHamilton: Good point. I'll update that right now.

bit-bender (approved):

PR #55, opened 2020-10-21
Refactor/cleanup source file names

PircDef:


tpm2/include/memory_util.h
```diff
@@ -1,12 +1,12 @@
 /**
- * @file  tpm2_kmyth_misc.h
+ * @file  memory_util.h
  *
  * @brief Provides miscellaneous utility functions for Kmyth applications
```
bit-bender, line 4: Should probably also change this comment from "miscellaneous utility functions" to "memory utility functions"

tpm2/src/util/memory_util.c
```diff
@@ -1,11 +1,11 @@
 /**
- * tpm2_kmyth_misc.c:
+ * memory_util.c:
  *
  * C library containing miscellaneous utilities supporting Kmyth applications
```
bit-bender, line 4: Part 2 of my comment for the memory utilities header file: In this comment also, "miscellaneous utilities" should probably be changed to something more descriptive, like "memory utilitiies"

PR #57, opened 2020-10-21
Organized some headers and removed duplicate declarations.

PircDef:
I also added some includes because some of the code depended on tss2 includes to bring in others libraries. I tried to minimize that dependence.

### DIFF
--- a/tpm2/include/tpm/formatting_tools.h
+++ b/tpm2/include/tpm/formatting_tools.h
@@ -8,10 +8,12 @@
 #ifndef FORMATTING_TOOLS_H
 #define FORMATTING_TOOLS_H
 
-#include <tss2/tss2_sys.h>
-#include "cipher/cipher.h"
 #include <stddef.h>
 #include <stdint.h>
+
+#include <tss2/tss2_sys.h>
+
+#include "cipher/cipher.h"
 
 typedef struct Ski_s
 {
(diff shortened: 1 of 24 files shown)


tpm2/include/tpm/formatting_tools.h
```diff
@@ -8,9 +8,10 @@
 #ifndef FORMATTING_TOOLS_H
 #define FORMATTING_TOOLS_H
 
-#include <stdlib.h>
 #include <tss2/tss2_sys.h>
```
b-carr, line 11: Reorder these: standard includes followed by non-standard (but system installed) includes followed by project includes.

tpm2/include/tpm/tpm2_interface.h
```diff
@@ -10,6 +10,8 @@
 
 #include <tss2/tss2_sys.h>
```
b-carr, line 11: Same suggestion here on the order of the includes.

b-carr (commented):
Rather than comment every place I'll make this comment here.

I'd prefer we order the includes
(1) Standard includes
(2) Non-standard includes not from our project
(3) Includes from our project.

PircDef:
I ordered the headers like this:

1. Corresponding header file
2. Standard includes
3. 3rd party includes
4. Headers from kmyth

Each is alphabetically sorted (starting with the directory where applicable)

PR #61, opened 2020-10-27
Add error checking and quality improvements to IO utilities

PeterHamilton:
This change adds a series of fixes to the IO utility functions found in file_io.c and formatting_tools.c, adding error checking on various internal and external library function calls, explicitly freeing unfreed memory during error conditions, setting unused and freed pointers to NULL, adding error logging where appropriate, along with additional minor fixes.

Partially addresses #25

### DIFF
--- a/tpm2/src/tpm/formatting_tools.c
+++ b/tpm2/src/tpm/formatting_tools.c
@@ -372,7 +372,9 @@ int parse_ski_bytes(uint8_t * input, size_t input_length, Ski * output)
                           &raw_pcr_select_list_data,
                           &raw_pcr_select_list_size,
                           KMYTH_DELIM_PCR_SELECTION_LIST,
-                          KMYTH_DELIM_STORAGE_KEY_PUBLIC))
+                          strlen(KMYTH_DELIM_PCR_SELECTION_LIST),
+                          KMYTH_DELIM_STORAGE_KEY_PUBLIC,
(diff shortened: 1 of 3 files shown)


tpm2/src/tpm/formatting_tools.c
```diff
@@ -829,27 +897,31 @@ int get_ski_block_bytes(char **contents,
                         uint8_t ** block, size_t *blocksize, char *delim,
                         char *next_delim)
 {
+  // TODO: Consider passing in the delim lengths as arguments and use strnlen
```
PircDef, line 900: I am in favor of passing in the delim lengths to get_ski_block_bytes

PeterHamilton: Ok, I'll make that change.

tpm2/src/util/file_io.c
```diff
@@ -118,28 +138,69 @@ int read_bytes_from_file(char *input_path, uint8_t ** data, size_t *data_length)
...
+              input_path);
+    if (!BIO_free(bio))
+    {
+      kmyth_log(LOG_ERR, "error freeing BIO ... exiting");
+    }
+    return 1;
+  }
   int input_size = st.st_size;
 
+  // TODO: What maximum file size should be supported here?
```
PircDef, line 164: I don't think there was ever an intended max file size. If the file is too large to fit in memory, the malloc should fail.

PR #66, opened 2020-11-05
Improvement/pcr handling interface

b-carr:


tpm2/include/kmyth.h
```diff
@@ -35,11 +35,12 @@
  *
  * @param[in]  oa_bytes_len      number of bytes in owner_auth_passwd
  *
- * @param[in]  pcrs_string       String indicating which PCRs, if any, to apply
+ * @param[in]  pcrs              Array containing PCRs values, if any, to apply
```
bit-bender, line 38: This is nit-picky, but I think this comment is potentially a little confusing. The array contains 'PCR selections' or 'PCR indices'. When I hear 'PCR values', I think of the contents of PCR registers, but perhaps the context makes it obvious that we are identifying PCRs not reading from them.

tpm2/src/main/seal.c
```diff
@@ -24,6 +24,105 @@
...
+      pcrs_array_size *= 2;
+    }
+    (*pcrs)[*pcrs_len] = (int) pcrIndex;
+    (*pcrs_len)++;
+    pcrs_string_cur = pcrs_string_next;
+    pcrs_string_next = NULL;
+  }
+
+  return 0;
+}
```
bit-bender, line 124: At first I got hung up on the fact that we were not validating the user's input values here beyond them being formatted correctly as a string of comma and/or space delimited integers (e.g., not checking that integers were in range of valid PCR selections, not checking for duplicates). After thinking about it, though, I think this is much cleaner and works well (e.g., invalid integers are already checked for and caught in init_pcr_selection(), duplicates are handled correctly without issue)

tpm2/src/tpm/kmyth_seal_unseal_impl.c
```diff
@@ -116,9 +117,9 @@ int tpm2_kmyth_seal(uint8_t * input,
   // will specify that no PCRs were selected by the user - all-zero mask)
   // This PCR Selection struct will be used in the authorization policy for
   // new, non-primary Kmyth objects.
-  if (init_pcr_selection(sapi_ctx, pcrs_string, &ski.pcr_list))
+  if (init_pcr_selection(sapi_ctx, pcrs, pcrs_len, &ski.pcr_list))
   {
-    kmyth_log(LOG_ERR, "error parsing PCR string: %s ... exiting", pcrs_string);
+    kmyth_log(LOG_ERR, "error intializing PCRs ... exiting");
```
bit-bender, line 122: "initializing" is mis-spelled

tpm2/src/tpm/pcrs.c
```diff
@@ -45,35 +46,28 @@ int init_pcr_selection(TSS2_SYS_CONTEXT * sapi_ctx,
...
-    bool *pcrs_list = NULL;
+    kmyth_log(LOG_DEBUG, "applying user-specified PCRs ...");
 
-    pcrs_list = malloc(numPCRs * sizeof(bool));
-    if (pcrs_list == NULL)
+    if (pcrs_len == 0)
     {
-      kmyth_log(LOG_ERR, "failed to allocate memory for PCR list ... exiting");
+      kmyth_log(LOG_ERR,
+                "non-NULL PRCs array supplied, but length is 0 ... exiting");
```
bit-bender, line 56: 'PCRs' is mis-spelled

PR #67, opened 2020-11-05
Update the Kmyth SGX Makefile to install libraries

### LINKED ISSUE #52: Add Install for libsgxseal
The kmyth sgx Makefile needs an install option that will install the shared objects and header alongside the tpm2 install. It also needs an uninstall to remove only the sgx portions.

PeterHamilton:
This change updates the Kmyth SGX Makefile, adding install and uninstall sections for the Kmyth SGX libraries and headers that place them alongside the existing Kmyth libraries and headers.

The install location of the signed seal library is now passed to the build process as a preprocessor directive, which is detected and used as the enclave path by the underlying source code.

Finally, the names of the Kmyth SGX libraries have been updated.

Closes #52

### DIFF
--- a/sgx/Makefile
+++ b/sgx/Makefile
@@ -30,6 +30,11 @@
 #
 #
 
+PREFIX ?= /usr/local
+INCLUDE_DIR ?= Include
+LIB_DIR ?= .
+INSTALL_LIB_PATH := $(DESTDIR)$(PREFIX)/lib
+
 ######## SGX SDK Settings ########
 
 SGX_SDK ?= /opt/intel/sgxsdk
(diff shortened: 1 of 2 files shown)


PeterHamilton:
This may require some additional debugging. The `demo` application was working but (for me) will now core dump due to `sgx_seal_get_sealed_size` returning a non-zero value.

I also am getting the following error/warning for both the `demo` application (before the core dump) and the `test.c` application:

```
Please use the correct uRTS library from PSW package.
```

I believe this is caused by linking against the SDK `libsgx_urts.so` library instead of the PSW version of this library. The Kmyth SGX Makefile however hard links against the `libsgx_urts.so` that is found under `/opt/intel/sgxsdk/lib64` which is obviously the SDK library. I'm not sure at this point if the library is ultimately causing the above core dump or if its the underlying PSW instance I have running erroring out in some way.

PircDef:
I think I was recreating this issue. I was getting:  

    demo: App/App.cpp:31: int main(int, const char**): Assertion `ret == 0' failed.

when I was running:

    $make
    $make install

Once I remembered to include the Make options in the README:

    $make SGX_MODE=SIM SGX_DEBUG=1 MITIGATION-CVE-2020-0551=NONE
    $make install

the demo began working. As it stands, the PR looks good to me, but I will wait to merge until you've looked at the issue further.

PeterHamilton:
Yep, I had totally forgotten about those `make` environment variables. Running the build with those, installing, and then testing the demo application works as expected.

I think this is good to go.

Issue #70, opened 2020-11-09
Add unit tests for kmyth TPM object tools


bit-bender:
Unit tests must be written and integrated into the kmyth unit testing framework for the kmyth TPM object tool functionality implemented in tpm2/src/tpm/object_tools.c.

PeterHamilton:
I'll start working on this one. There's some low hanging fruit that should be easy to add tests for.

PR #82, opened 2020-11-17
Add unit tests for some of the TPM object tools utility functions

PeterHamilton:
This change adds unit tests for some of the simpler TPM 2.0 object tools and utility functions, focusing mostly on initializing various objects and structures. A new unit test suite has been added to organize the new tests and the relevant headers and the project Makefile have also been updated to incorporate the new functionality.

Partially addresses #70


tpm2/test/tpm/object_tools_test.c
```diff
@@ -0,0 +1,236 @@
...
+  CU_ASSERT(sensitiveArea.size == 4);
+
+  // Non-empty object auth and data should yield a valid sensitive area
+  object_auth.size = 4;
+  uint8_t object_data[] = {1, 2, 3, 4};
+  size_t object_dataSize = 4;
+
+  CU_ASSERT(init_kmyth_object_sensitive(object_auth, object_data,
+                                        object_dataSize, &sensitiveArea) == 0);
+  CU_ASSERT(sensitiveArea.sensitive.userAuth.size == 4);
```
bit-bender, line 78: This is kind of a trivial comment, but since we setup a parameterized test case here, should we use the variables in the assertions? That way, we could potentially alter the test case definition without affecting the test itself. In my subjective opinion, using the variable names also makes it slightly more understandable:
- CU_ASSERT(sensitiveArea.sensitive.userAuth.size == object_auth.size)
- CU_ASSERT(sensitiveArea.data.size == object_dataSize)
- CU_ASSERT(sensitiveArea.size = object_auth.size + object_dataSize + 4)

bit-bender (commented):
I decided to leave one non-substantive comment. Please review and, if you agree, please make the suggested minor changes. If you disagree or believe that they are not worth the time to pursue, feel free to let me know that as well.

PeterHamilton:
@bit-bender Yep, I think that makes sense. I'll update those asserts and will push that soonish.

tpm2/test/tpm/object_tools_test.c
PeterHamilton: Done

PR #83, opened 2020-11-18
Unit Tests for AES/GCM cipher

bit-bender:
Added unit tests for the AES/GCM cipher functionality implemented in aes_gcm_encrypt() and aes_gcm_decrypt() contained in tpm2/src/cipher/aes_gcm.c. Also updated the makefile to build unit tests in the tpm2/test/cipher directory. Finally, I made two simple updates to log messages in the kmyth AES/GCM cipher code that I noticed were incomplete or wrong.


tpm2/test/cipher/aes_gcm_test.c
```diff
@@ -0,0 +1,307 @@
...
+  size_t plaintext_len = 16;
+  size_t ciphertext_len = 0;
+  size_t decrypt_len = 0;
+  
+  key = calloc(key_len, 1);
+  plaintext = calloc(plaintext_len, 1);
+  
+  CU_ASSERT(aes_gcm_encrypt(key, key_len, plaintext,
+            plaintext_len, &ciphertext, &ciphertext_len) == 0);
+  CU_ASSERT(ciphertext_len = plaintext_len + GCM_IV_LEN + GCM_TAG_LEN);
```
PircDef, line 79: I think this should have been "=="

tpm2/test/cipher/aes_gcm_test.c
```diff
@@ -0,0 +1,307 @@
...
+  // alter the last byte of the tag and verify decryption failure
+  ciphertext[ciphertext_len-1] ^= 0x1;
+  CU_ASSERT(aes_gcm_decrypt(key, key_len, ciphertext, ciphertext_len,
+                            &decrypt, &decrypt_len) == 1);
+  decrypt_len = 0;
+  decrypt = NULL;
+ 
+  // truncate the tag by 2 bytes and verify decryption failure
+  ciphertext_len -= 2;
+  CU_ASSERT(aes_gcm_decrypt(key, key_len, ciphertext, ciphertext_len,
```
PircDef, line 164: If somehow the length wasn't failing, wouldn't this fail anyway because of:

    ciphertext[ciphertext_len-1] ^= 0x1;

tpm2/test/cipher/aes_gcm_test.c
```diff
@@ -0,0 +1,307 @@
...
+  // alter the first byte of IV and verify decryption failure
+  ciphertext[0] ^= 0x1;
+  CU_ASSERT(aes_gcm_decrypt(key, key_len, ciphertext, ciphertext_len,
+                            &decrypt, &decrypt_len) == 1);
+  decrypt = NULL;
+  decrypt_len = 0;
+
+  // truncate the IV and verify decryption failure
+  unsigned char* truncated_iv_cipher = ciphertext + 2;
+  CU_ASSERT(aes_gcm_decrypt(key, key_len, truncated_iv_cipher, ciphertext_len-2,
```
PircDef, line 208: Similar to above, won't this one also already fail without the iv modification because:

    ciphertext[0] ^= 0x1;

tpm2/test/cipher/aes_gcm_test.c
bit-bender: Yes, good catch! I corrected this.

tpm2/test/cipher/aes_gcm_test.c
bit-bender: I think that the modified byte (last byte of ciphertext array is the last byte of tag) is one of the two bytes that is truncated by shortening the length, so it shouldn't fail because of that. Nonetheless, I think it probably makes sense to test the two more independently. Therefore, I reversed the order of the tests and only decremented the length (to truncate the tag) in the aes_gcm_decrypt() call itself.

tpm2/test/cipher/aes_gcm_test.c
bit-bender: Again, the modified byte (first byte ciphertext array is first byte of IV) is one that is not passed to aes_gcm_decrypt() in the truncated IV case (ciphertext array starts at byte two past the real first byte). Again, though, I think it made sense to reverse the order here to make the two cases more independent.

PircDef (approved):

Issue #71, opened 2020-11-09
Add unit tests for kmyth PCR utilities


bit-bender:
Unit tests must be written and integrated into the kmyth unit testing framework for the kmyth PCR utility functionality implemented in tpm2/src/tpm/pcrs.c.

PR #88, opened 2020-12-08
Added unit tests for pcrs.c

PircDef:
Added unit tests for pcrs.c. Addresses #71

### DIFF
--- a/tpm2/test/src/tpm/pcrs_test.c
+++ b/tpm2/test/src/tpm/pcrs_test.c
@@ -95,7 +95,7 @@ void test_get_pcr_count(void)
 	int count = 0;
 	//Valid get count
 	CU_ASSERT(get_pcr_count(sapi_ctx, &count) == 0);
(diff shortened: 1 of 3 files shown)


tpm2/test/src/tpm/pcrs_test.c
```diff
@@ -0,0 +1,102 @@
...
+	bool emulator = true;
+	get_tpm2_impl_type(sapi_ctx, &emulator);
+	if(!emulator)
+	{
+		return;
+	}
+
+	int count = 0;
+	//Valid get count
+	CU_ASSERT(get_pcr_count(sapi_ctx, &count) == 0);
```
b-carr, line 97: I suggest adding a comment here that since we don't know how many PCRs should be found, the best we can do is test that the result is non-zero. Ideally we would know the right answer and test explicitly, but the level of effort to get that setup probably exceeds the value it would add.

b-carr (approved):

Issue #94, opened 2020-12-11
Make get_srk_handle() more readable/modular


bit-bender:
The function get_srk_handle() currently gets a list of all objects in TPM persistent storage. It then has a for loop that calls check_if_srk() on each of those objects. As soon as it comes across an object in that list that passes all of the criteria tested in check_if_srk() (i.e., check_if_srk() sets the isSRK boolean to true), it sets srk_handle to that handle value and breaks out of the for loop. Since srk_handle will be non-zero in this case, get_srk_handle() will just exit normally at this point with the srk_handle parameter available to the calling function. If the for loop exits after check_if_srk() has failed on all loaded objects, srk_handle will have never been changed from its zero initialization. In this case, derive_srk() will be called to re-derive the SRK from its seed and load it into persistent TPM storage. In this case derive_srk() provides the handle referencing the SRK in TPM storage to get_srk_handle() so that it can provide it to its calling function.

While the logic in get_srk_handle() seems fine, this issue attempts to make the function more readable and the code a little more modular. After discussion about this, a decision to simplify get_srk_handle() by abstracting it into the following two function calls, seemed the best way to achieve this:

- get_existing_srk_handle() gets the list of objects in TPM persistent storage and checks them all against the SRK criteria. Upon return from this function, the srk_handle parameter passed to it is either zero (none of the loaded objects is the SRK) or contains the handle used to reference the SRK.
- put_srk_into_tpm_persistent_storage() is a renamed derive_srk() function.

The new get_srk_handle() would call get_existing_srk_handle(). It would then check the srk_handle parameter resulting from this call and call put_srk_into_tpm_persistent_storage() if srk_handle is still zero (i.e., wasn't found by get_existing_srk_handle()).

PR #95, opened 2020-12-11
More readable and modular get_srk_handle()

bit-bender:
This pull request proposes no real change to the underlying logic in get_srk_handle() but adds some abstraction in an attempt to make the get_srk_handle() function more readable and the code a little more modular.

Specifically, as a result of this proposed change, get_srk_handle() calls get_existing_srk_handle(), a new function where much of the get_srk_handle() code was copied and pasted.. It then checks the srk_handle parameter resulting from the call to get_existing_srk_handle(). if srk_handle is still zero, the SRK was not found already loaded into the TPM and put_srk_into_persistent_storage() to reestablish the SRK within the TPM. put_srk_into_persistent() storage is basically a renamed derive_srk().

This pull request addresses issue #94.

### DIFF
--- a/tpm2/src/tpm/storage_key_tools.c
+++ b/tpm2/src/tpm/storage_key_tools.c
@@ -33,7 +33,11 @@ int get_srk_handle(TSS2_SYS_CONTEXT * sapi_ctx,
   // available TPM persistent storage location where the SRK can be put.
   TPM2_HANDLE next_persistent_handle = 0;
 
-  get_existing_srk_handle(sapi_ctx, srk_handle, &next_persistent_handle);
+  if(get_existing_srk_handle(sapi_ctx, srk_handle, &next_persistent_handle))
+  {
+    kmyth_log(LOG_ERR, "error retrieving SRK handle from TPM ... exiting");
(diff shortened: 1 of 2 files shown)


tpm2/src/tpm/storage_key_tools.c
```diff
@@ -23,12 +23,48 @@
...
+  // Get the list of objects in TPM persistent storage and check them all
+  // against the SRK criteria. Upon return from get_existing_srk_handle(),
+  // the srk_handle parameter passed to get_existing_srk_handle() is either
+  // zero (none of the loaded objects is the SRK) or contains the handle used
+  // to reference the SRK in TPM persistent storage. The function
+  // get_existing_srk_handle() will set next_persistent_handle to the next
+  // available TPM persistent storage location where the SRK can be put.
+  TPM2_HANDLE next_persistent_handle = 0;
+
+  get_existing_srk_handle(sapi_ctx, srk_handle, &next_persistent_handle);
```
PircDef, line 36: This needs to check for error conditions, e.g. it returns 1 if sapi_ctx is null.

Issue #72, opened 2020-11-09
Add unit tests for kmyth TPM storage key tools


bit-bender:
Unit tests must be written and integrated into the kmyth unit testing framework for the kmyth utility functions that manage and interact with TPM keys (i.e., storage root key and storage keys). More specifically, add tests for the functionality implemented in tpm2/src/tpm/storage_key_tools.c.

PR #96, opened 2020-12-16
Tests/add storage key tools tests

PircDef:
Addresses #72

### DIFF
--- a/tpm2/test/src/tpm/storage_key_tools_test.c
+++ b/tpm2/test/src/tpm/storage_key_tools_test.c
@@ -121,7 +121,7 @@ void test_check_if_srk(void)
 	create_policy_digest(sapi_ctx, pcrs_struct, &auth_policy);
 	TPM2B_PRIVATE sk_priv = {.size = 0,};
 	TPM2B_PUBLIC sk_pub = {.size = 0,};
-  TPM2_HANDLE sk_handle = 0;
+	TPM2_HANDLE sk_handle = 0;
 	create_and_load_sk(sapi_ctx, srk_handle, owner_auth, obj_auth, pcrs_struct, auth_policy, &sk_handle, &sk_priv, &sk_pub);
 	CU_ASSERT(check_if_srk(sapi_ctx, sk_handle, &is_srk) == 0);
 	CU_ASSERT(!is_srk);
--- a/tpm2/test/src/tpm/storage_key_tools_test.c
+++ b/tpm2/test/src/tpm/storage_key_tools_test.c
@@ -154,7 +154,9 @@ void test_put_srk_into_persistent_storage(void)
 	get_existing_srk_handle(sapi_ctx, &srk_handle, &next);
 	TPM2_HANDLE old_srk = srk_handle;
 	srk_handle = next;
-	//Test for failure if the key is already in storage
+	//Test for failure if we try to load at a location that is already in use
+	CU_ASSERT(put_srk_into_persistent_storage(sapi_ctx, old_srk, auth) != 0);
+	//Loading second copy of SRK at next available persistent handle should work
 	CU_ASSERT(put_srk_into_persistent_storage(sapi_ctx, srk_handle, auth) == 0);
 	TPM2B_AUTH emptyAuth = { .size = 0 };
 	TPM2B_NONCE emptyNonce = { .size = 0 };
@@ -168,7 +170,7 @@ void test_put_srk_into_persistent_storage(void)
                                                  .hmac = emptyAuth
                                               }}
                                    };
-	//Clear the loaded srk
+	//Clear all persistent storage to remove SRK and make TPM2_PERSISTENT_FIRST available
 	Tss2_Sys_Clear(sapi_ctx, TPM2_RH_PLATFORM, &cmdAuth, &cmdRsp);
 	next = 0;
 	srk_handle = 0;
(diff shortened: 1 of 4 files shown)


tpm2/test/src/tpm/storage_key_tools_test.c
```diff
@@ -0,0 +1,200 @@
...
+void test_put_srk_into_persistent_storage(void)
+{
+	TSS2_SYS_CONTEXT *sapi_ctx = NULL;
+	init_tpm2_connection(&sapi_ctx);
+	TPM2B_AUTH auth = {.size=0,};
+
+	//NULL context
+	CU_ASSERT(put_srk_into_persistent_storage(NULL, 0, auth) != 0);
+
+	//Valid test
```
bit-bender, line 150: Unless this gets run after the TPM is cleared, the SRK will most likely already be loaded and get_existing_srk_handle() will likely produce a non-zero srk_handle (e.g., 0x81000000). Thus, the CU_ASSERT() will never be run. As it will have been run successfully prior to this, however, in order to create this initial condition, it is valid to argue that this is indirectly tested. If you want to actually exercise the function and test its return value here, we could just also load the SRK into the next persistent handle (if it is already loaded, this will load the SRK at a second persistent handle). As repeated testing could fill up persistent storage, though, some cleanup may be required. Alternatively the TPM could be cleared before running the test, always loading it at the first persistent handle. Finally, the test could be left as is, using the argument that it is OK that this CU_ASSERT() does not directly test the function. Below is some code I used to play around with this:

```
//Valid test - load SRK into next available persistent handle, clear TPM
TPM2_HANDLE next = 0;
TPM2_HANDLE srk_handle = 0;
get_existing_srk_handle(sapi_ctx, &srk_handle, &next);
printf("Before put: srk_handle = 0x%08x, next = 0x%08x\n", srk_handle, next);
srk_handle = next;
CU_ASSERT(put_srk_into_persistent_storage(sapi_ctx, srk_handle, auth) == 0);
get_existing_srk_handle(sapi_ctx, &srk_handle, &next);
printf("After put: srk_handle = 0x%08x, next = 0x%08x\n", srk_handle, next);
TPM2B_AUTH emptyAuth = { .size = 0 };
TPM2B_NONCE emptyNonce = { .size = 0 };
TSS2L_SYS_AUTH_RESPONSE cmdRsp;
TSS2L_SYS_AUTH_COMMAND cmdAuth = {
                                     .count = 1,
                                     .auths = {{
                                                .sessionHandle = TPM2_RS_PW,
                                                .sessionAttributes = 0,
                                                .nonce = emptyNonce,
                                                .hmac = emptyAuth
                                              }}
                                   };
uint32_t rc = Tss2_Sys_Clear(sapi_ctx, TPM2_RH_PLATFORM, &cmdAuth, &cmdRsp);
printf("Tss2_Sys_Clear: rc = 0x%08x\n", rc);
get_existing_srk_handle(sapi_ctx, &srk_handle, &next);
printf("After Clear: srk_handle = 0x%08x, next = 0x%08x\n", srk_handle, next);
```

PircDef: I like this fix. I'll get a new commit up shortly.

bit-bender: I think that either this comment is confusing or this is not what you intended:

```
//Test for failure if the key is already in storage
CU_ASSERT(put_srk_into_persistent_storage(sapi_ctx, srk_handle, auth) == 0);
```

I think one or both of the two following comment/assertion pairs are the actual possibilities (first is a changed assertion, second is same assertion with a reworded comment):

```
//Test for failure if we try to load at a location that is already in use
CU_ASSERT(put_srk_into_persistent_storage(sapi_ctx, old_srk, auth) != 0);
//Loading second copy of SRK at next available persistent handle should work
CU_ASSERT(put_srk_into_persistent_storage(sapi_ctx, srk_handle, auth) == 0);
```

This is not technically incorrect, but the comment for the following code snippet implies that you are clearing a single persistent handle:

```
//Clear the loaded srk
Tss2_Sys_Clear(sapi_ctx, TPM2_RH_PLATFORM, &cmdAuth, &cmdRsp);
```
In fact, with the current code, I believe that you have the original SRK loaded at the first persistent storage handle (0x81000000) and a second copy of the SRK loaded at the second persistent storage handle (0x81000001). Therefore, I believe the following comment is a little more clear:

```
//Clear all persistent storage
Tss2_Sys_Clear(sapi_ctx, TPM2_RH_PLATFORM, &cmdAuth, &cmdRsp);
```
Finally, I believe that the following assertion will not work in all cases:

```
//Verify it has loaded correctly
CU_ASSERT(srk_handle == old_srk);
```

I do not believe that you can count on the initial conditions always being the TPM with the SRK loaded into the first persistent storage handle. The TPM could technically have some other (non-SRK) object loaded at the first handle in TPM storage. For kmyth, the only thing we load into the TPM's persistent storage is the SRK, but if the TPM was being used by some other application, it might load something into persistent storage before kmyth loads the SRK. Then the SRK would not be at the first persistant storage handle and old_srk would not be set to the first persistent storage handle. The Tss2_Sys_Clear() would remove all objects from persistent storage, so get_existing_srk_handle() should set next to the first persistent storage handle (0x81000000). I believe the following version would always be correct:

```
//Verify it has loaded at the first persistent storage handle
CU_ASSERT(srk_handle == 0x81000000);
```

bit-bender (approved):

PR #100, opened 2021-01-13
Added tests applying AES/GCM decryption test vectors

bit-bender:
Added tests to apply applicable NIST AES/GCM decryption vectors to the kmyth AES/GCM decrypt cipher implementation (i.e., those with IV and tag implementations that match those used).

As part of this effort, it was discovered that kmyth implementation failed for vectors where the ciphertext input was the encryption of an empty plaintext (i.e, the decryption result should return a valid IV and tag, but an empty plaintext result). This pull request, therefore, also fixes this issue.

Note: Although, I did not change the configuration for the indent utility in the makefile, a number of white-space edits were generated by building the code, and included in this very simple pull request.

### DIFF
--- a/tpm2/include/kmyth.h
+++ b/tpm2/include/kmyth.h
@@ -49,7 +49,7 @@
  * @return 0 on success, 1 on error
  */
 int tpm2_kmyth_seal(uint8_t * input, size_t input_len,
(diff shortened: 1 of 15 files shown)


bit-bender:
Replaced by PR #104

PR #104, opened 2021-01-14
Added tests applying AES/GCM decryption test vectors

bit-bender:
Added tests to apply applicable NIST AES/GCM decryption vectors to the kmyth AES/GCM decrypt cipher implementation (i.e., those with IV and tag implementations that match those used).

As part of this effort, it was discovered that kmyth implementation failed for vectors where the ciphertext input was the encryption of an empty plaintext (i.e, the decryption result should return a valid IV and tag, but an empty plaintext result). This pull request, therefore, also fixes this issue.

### DIFF
--- a/tpm2/test/src/cipher/aes_gcm_test.c
+++ b/tpm2/test/src/cipher/aes_gcm_test.c
@@ -130,7 +130,6 @@ int get_aes_gcm_vector_from_file(FILE * fid,
         {
           key_str[--key_str_len] = '\0';  // strip any trailing '\n' or 'r'
         }
-        Key_len = key_str_len / 2;  // 2 hex chars map to a byte of key
         convert_HexString_to_ByteArray((char **) &Key,
                                        key_str,
                                        key_str_len);
--- a/tpm2/test/src/cipher/aes_gcm_test.c
+++ b/tpm2/test/src/cipher/aes_gcm_test.c
@@ -329,12 +329,6 @@ void test_aes_gcm_decrypt_vectors(void)
   // array of file pointers for test vector files
   FILE * test_vector_fd[MAX_VECTOR_SETS_IN_COMPILATION] = {NULL};
 
-  // counter to track number of test vectors applied from a file
-  int test_vector_count = 0;
-
-  // flag used to signal end of processing for a given test vector file
-  bool done_with_test_vector_file = false;
-
   // check that number of test vector files complies with specified maximum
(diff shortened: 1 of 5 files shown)


tpm2/test/src/cipher/aes_gcm_test.c
```diff
@@ -56,6 +63,411 @@ int aes_gcm_add_tests(CU_pSuite suite)
...
+        .func_to_test = "aes_gcm_decrypt",
+        .path = "./test/data/gcmtestvectors/gcmDecrypt256.rsp" }
+    }
+  };
+
+  // array of file pointers for test vector files
+  FILE * test_vector_fd[MAX_VECTOR_SETS_IN_COMPILATION] = {NULL};
+
+  // counter to track number of test vectors applied from a file
+  int test_vector_count = 0;
```
b-carr, line 333: I think it's cleaner if this and `done_with_test_vector_file` are local to the loop over the files.

tpm2/test/src/cipher/aes_gcm_test.c
```diff
@@ -56,6 +63,411 @@ int aes_gcm_add_tests(CU_pSuite suite)
...
+                ((key_str[key_str_len-1] == '\n') ||
+                 (key_str[key_str_len-1] == '\r')) )
+        {
+          key_str[--key_str_len] = '\0';  // strip any trailing '\n' or 'r'
+        }
+        Key_len = key_str_len / 2;  // 2 hex chars map to a byte of key
+        convert_HexString_to_ByteArray((char **) &Key,
+                                       key_str,
+                                       key_str_len);
+        Key_len = key_str_len / 2;  // 2 hex chars map to a byte of key
```
b-carr, line 137: I don't see any reason to do this twice (lines 133 and 137), but in case it's important I didn't change it when I was editing the comments.

tpm2/test/src/cipher/aes_gcm_test.c
bit-bender: I moved the declaration/initialization for test_vector_count and done_with_test_vector_file right before the while loop iteratively processing the file.

tpm2/test/src/cipher/aes_gcm_test.c
bit-bender: This must have been a copy/paste remnant that I didn't clean up properly. I deleted the redundant assignment.

PR #105, opened 2021-01-14
Add Tests for AES Key Wrap Cipher Functionality

bit-bender:
Added tests to apply NIST AES Key Wrap (RFC-3394 and RFC-5649) vectors to the kmyth AES Key Wrap cipher implementation.

### DIFF
--- a/tpm2/test/include/cipher/kmyth_cipher_test.h
+++ b/tpm2/test/include/cipher/kmyth_cipher_test.h
@@ -1,0 +1,62 @@
+/**
+ * @file  kmyth_cipher_test.h
+ */
+
+#ifndef KMYTH_CIPHER_TEST_H
+#define KMYTH_CIPHER_TEST_H
+
+/**
+ * Specify maximum number of test vector sets (vector files) that can be
+ * contained within a "vector set compilation" (used to size that array).
+ */
+#define MAX_VECTOR_SETS_IN_COMPILATION 20
+
+/**
+ * Specify maximum number of test vectors to process when parsing
+ * a test vector file.
+ */
+#define MAX_KEYWRAP_TEST_VECTOR_COUNT 500
+#define MAX_GCM_TEST_VECTOR_COUNT 7875
+
+/**
+ * Specify the maximum length (in chars) of a test vector component
+ * This is needed to appropriately size the buffers used to parse and
+ * process test vector components read from a file. For example, a
+ * 2048 hexadecimal character string can specify a 512-byte or
+ * 4096-bit test vector component.
+ */
+#define MAX_TEST_VECTOR_COMPONENT_LENGTH 2500
+
+typedef struct cipher_vector_set
+{
+  char * desc;
+  char * func_to_test;
+  char * path;
+} cipher_vector_set;
+
+typedef struct cipher_vector_compilation
+{
+  size_t count;
+  cipher_vector_set sets[MAX_VECTOR_SETS_IN_COMPILATION];
+} cipher_vector_compilation;
+
+
+/**
+ * As the NIST test vectors are specified as hexadecimal values, the
+ * bytes encrypted or decrypted by the kmyth keywrap cipher
+ * implementation must be converted into a hex format for comparison
+ * with the expected result. This simple utility provides that
+ * functionality.
+ *
+ * @param[out] result  - Byte array corresponding to input hex string value
+ *
+ * @param[in]  hex_str - hexadecimal string to be converted
+ *
+ * @param[in]  size    - length (in hex chars) of input string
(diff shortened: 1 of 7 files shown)


tpm2/test/include/cipher/kmyth_cipher_test.h
```diff
@@ -0,0 +1,63 @@
...
+
+typedef struct cipher_vector_compilation
+{
+  size_t count;
+  cipher_vector_set sets[MAX_VECTOR_SETS_IN_COMPILATION];
+} cipher_vector_compilation;
+
+
+/**
+ * As the NIST test vectors are specified as hexadecimal values, the
```
b-carr, line 46: I think the comment is backwards, this function converts the hex provided by NIST to bytes, rather than the bytes provided by kmyth to hex.

tpm2/test/src/cipher/aes_keywrap_test.c
```diff
@@ -0,0 +1,539 @@
...
+  size_t inData_len = 0;
+
+  unsigned char* outData = NULL;
+  size_t outData_len = 0;
+
+  // Test failure on null key
+  inData = malloc(16);
+  inData_len = 16;
+  key_len = 16;
+  CU_ASSERT(aes_keywrap_3394nopad_encrypt(key, key_len, inData, inData_len, &outData, &outData_len) == 1);
```
b-carr, line 245: I think after these failure checks you can also verify that `outData == NULL` and that `outDataLen == 0`. The failure conditions in encrypt/decrypt don't always `NULL` `outData`, but the input sanity checks happen before it gets allocated. This applies to all the tests in this function.

tpm2/test/src/cipher/aes_keywrap_test.c
```diff
@@ -0,0 +1,539 @@
...
+
+      // Done with the test vector file (processed all vectors or reached max)
+      fclose(test_vector_fd[i]);
+
+      // Provide INFO: message indicating how many test vectors were applied
+      printf("INFO: %s - %d test vectors applied\n",
+              aes_keywrap_vectors.sets[i].path, test_vector_count);
+
+      // reset flag/counters for processing of next (if any) test vector file
+      done_with_test_vector_file = false;
```
b-carr, line 522: I think it would be cleaner to have these variables local to the `i` loop and set at the start of each iteration, rather than having to reset them at the end of each iteration.

tpm2/test/include/cipher/kmyth_cipher_test.h
bit-bender: Yes, this was backwards. I attempted to rewrite the description of convert_HexString_to_ByteArray() in a manner that is hopefully more correct and less confusing.

tpm2/test/src/cipher/aes_keywrap_test.c
bit-bender: I think that I added a "CU_ASSERT((outData == NULL) && (outData_len == 0))" statement after all of the tests for out-of-limit input parameter handling.

tpm2/test/src/cipher/aes_keywrap_test.c
bit-bender: I moved the test_vector_count and done_with_test_vector_file variable declaration and initialization to just before the while loop used to process a test vector file ... much cleaner. In looking over this code, though, I noticed a number of other things that I thought might also improve it.

I also moved the check for (test_vector_count > MAX_KEYWRAP_TEST_VECTOR_COUNT) inside the while loop and set done_with_test_vector_file true, when exceeded. This simplifies the while condition to (!done_with_test_vector_file).

I moved the declaration and initialization of vector_passed to immediately precede the nested if statements that use it to created a boolean that indicates if all of the checks for that test vector passed, so that a single CU_ASSERT() can be used to report pass/fail status for that vector. Previously, it was declared and initialized much earlier in the while loop code block.

I also removed some printf() statements that I think were debug remnants from when I was originally writing and testing this code.

Finally, although I don't think it should ever happen, I added an else condition with a CU_FAIL() message if the test vector application results in a return code of -1. I believe that any of the kmyth cipher functions that these vectors are applied to should return only 0 if successful or 1 if an error is encountered. As we initialize rc to -1 and then assign it to the return value of the function call represented by our test vector, it should remain -1 only if the return code assignment does not actually happen or the called function returns -1 unexpectedly.

bit-bender:
Latest commit updates to a hopefully less obtrusive means of reporting to the user whether the unit tests include test-vector-based tests or not. Lastest commit also attempts to implement some of the recent changes to the vector file processing loop for the AES/GCM code in a similar manner.

PR #106, opened 2021-01-21
Test/kmyth seal unseal impl

b-carr:
This merge adds tests for the kmyth seal/unseal functionality.

### DIFF
--- a/tpm2/test/src/kmyth-test.c
+++ b/tpm2/test/src/kmyth-test.c
@@ -22,6 +22,7 @@
 #include "tpm2_interface_test.h"
 #include "storage_key_tools_test.h"
 #include "pcrs_test.h"
+#include "kmyth_seal_unseal_imlp_test.h"
 
 /**
  * Use trivial (do nothing) init_suite and clean_suite functionality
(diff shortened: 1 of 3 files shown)


tpm2/test/src/kmyth-test.c
```diff
@@ -71,51 +72,51 @@ int main(int argc, char** argv)
...
+    {
+      CU_cleanup_registry();
+      return CU_get_error();
+    }
+  if (storage_key_tools_add_tests(storage_key_tools_test_suite))
+    {
+      CU_cleanup_registry();
+      return CU_get_error();
+    }
+
```
bit-bender, line 89: In reviewing the changes for this file, I noticed that the additions use a style that is different than that used in the parts that were original. While I do not have a significant preference for one style over the other, we attempt to utilize the indent utility on the source code that is not in the test directory to keep our style consistent. In response to this, I submitted pull request #107 in an attempt to address this (and found out that we were not running indent on many of the non-test header files as well).

tpm2/test/src/tpm/kmyth_seal_unseal_impl_test.c
```diff
@@ -0,0 +1,287 @@
...
+  CU_ASSERT(output == NULL);
+  CU_ASSERT(output_len == 0);
+
+  // Check that if all inputs are valid seal produces correct (or at least) unsealable output of the right
+  // length.
+  // Note that we can't hard-code the exact output because we can't control what wrapping key is used.
+  CU_ASSERT(tpm2_kmyth_seal(input, input_len, &output, &output_len, auth_bytes, auth_bytes_len, owner_auth_bytes, oa_bytes_len, pcrs, pcrs_len, NULL) == 0);
+
+  size_t expected_output_len = 1569;
+  CU_ASSERT(output_len == expected_output_len);
```
bit-bender, line 106: This assertion (output_len == 1569) fails for me. In my test, output_len = 2259. I haven't looked into why this is and what makes sense, yet.

tpm2/test/src/kmyth-test.c
b-carr: I ran indent on the file which should have fixed the formatting (but didn't do the other test files since you already put in a pull request for those.)

tpm2/test/src/tpm/kmyth_seal_unseal_impl_test.c
b-carr: I think the right answer is to remove that ASSERT (which I've done.) The exact length of the sealed blob depends on TPM internals, so it doesn't really make sense for us to test that. I've left the test that the sealed blob unseals correctly.

PR #111, opened 2021-02-10
added option to build kmyth logger features into separate shared library

bit-bender:
Added "logger-lib" target to the Makefile, so that a shared library containing only the logger functionality can be optionally built. This is not a requirement for kmyth, but provides a way to build a kmyth-logger library so that the logger functionality can be exported, potentially, to other applications.

Moved build directory creation out of the "pre" target and implemented instead with makefile rules.

Moved a few things around in an attempt to make the grouping more logical.

Re-implemented the "delete back-ups" (i.e., file names ending in tilde) with a call to find.

Added a short blurb about how to build the "logger only" library to the INSTALL.md file.

### DIFF
--- a/tpm2/Makefile
+++ b/tpm2/Makefile
@@ -478,7 +478,11 @@ endif
 uninstall:
 	rm -f $(DESTDIR)$(PREFIX)/lib/$(TPM_UTIL_LIB_SONAME)
 	rm -f $(DESTDIR)$(PREFIX)/lib/$(LOGGER_LIB_SONAME)
+	rm -f $(DESTDIR)$(PREFIX)/include/kmyth/kmyth.h
+	rm -f $(DESTDIR)$(PREFIX)/include/kmyth/kmyth_log.h
+ifeq ($(wildcard $(DESTDIR)$(PREFIX)/include/kmyth/*.h),)
 	rm -rf $(DESTDIR)$(PREFIX)/include/kmyth
+endif
 	rm -f $(DESTDIR)$(PREFIX)/bin/kmyth-seal
 	rm -f $(DESTDIR)$(PREFIX)/bin/kmyth-unseal
 
--- a/tpm2/Makefile
+++ b/tpm2/Makefile
@@ -297,7 +297,6 @@ $(LIB_DIR)/libkmyth-tpm.so: $(UTIL_OBJECTS) \
 
 
 $(BIN_DIR)/kmyth-seal: $(MAIN_OBJ_DIR)/seal.o \
-                       $(LIB_DIR)/libkmyth-logger.so \
                        $(LIB_DIR)/libkmyth-tpm.so | \
                        $(BIN_DIR)
 	$(CC) $(MAIN_OBJ_DIR)/seal.o \
@@ -308,7 +307,6 @@ $(BIN_DIR)/kmyth-seal: $(MAIN_OBJ_DIR)/seal.o \
 	      -lkmyth-tpm
 
 $(BIN_DIR)/kmyth-unseal: $(MAIN_OBJ_DIR)/unseal.o \
-                         $(LIB_DIR)/libkmyth-logger.so \
                          $(LIB_DIR)/libkmyth-tpm.so | \
 												 $(BIN_DIR)
 	$(CC) $(MAIN_OBJ_DIR)/unseal.o \
@@ -319,7 +317,6 @@ $(BIN_DIR)/kmyth-unseal: $(MAIN_OBJ_DIR)/unseal.o \
 	      -lkmyth-tpm
 
 $(BIN_DIR)/kmyth-getkey: $(MAIN_OBJ_DIR)/getkey.o \
-                         $(LIB_DIR)/libkmyth-logger.so \
                          $(LIB_DIR)/libkmyth-tpm.so | \
                          $(BIN_DIR)
 	$(CC) $(MAIN_OBJ_DIR)/getkey.o \
@@ -410,7 +407,9 @@ pretest:
 	indent $(INDENT_OPTS) $(TEST_HEADERS)
 
 
-$(BIN_DIR)/kmyth-test: $(TEST_OBJECTS) | $(BIN_DIR)
+$(BIN_DIR)/kmyth-test: $(TEST_OBJECTS) \
+                       $(LIB_DIR)/libkmyth-tpm.so | \
+                       $(BIN_DIR)
 	$(CC) $(TEST_OBJECTS) \
 	      -o $(BIN_DIR)/kmyth-test \
 	      $(LDFLAGS) \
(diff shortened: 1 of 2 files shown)


bit-bender:
This pull request adds the ability to build/install a kmyth-logger library that contains only the logging features. Originally, this was just an additional option, either in addition to or in lieu of the kmyth-tpm library that incorporated both the TPM utility and logging features.

With the most recent commit, the build process instead splits the functionality across the two libraries. Thus, the TPM-based functionality in the kmyth-tpm library requires the logging functionality in the kmyth-logger library. There are now basically three build options:
1. *make* or *make all* will build both libraries and all three applications
2. *make logger-lib* will build only the kmyth-logger library
3. *make tpm-util-lib* will build both libraries, but not the applications (e.g., kmyth-seal, kmyth-unseal, kmyth-getkey)

The *install* target (i.e., running *sudo make install*) now includes conditional logic to only install the files that correspond to what has been built.

The INSTALL.md file has be updated to reflect this new approach.

b-carr (commented):
Using `make tpm-util-lib` doesn't create the `bin/` directory, which causes `make test` to fail. I don't know if we should have `make test` create `bin` or move the test executable to a different directory that `make test` creates if necessary, but since the tests only rely on the library code I think they should work if you've only build the library.

tpm2/Makefile
```diff
@@ -386,65 +402,113 @@ kmyth-test: $(TEST_OBJECTS)
...
+endif
 
 .PHONY: uninstall
 uninstall:
-	rm -f $(DESTDIR)$(PREFIX)/lib/$(KMYTH_LIB_SONAME)
-	rm -f $(DESTDIR)$(PREFIX)/include/kmyth/kmyth_log.h
-	rm -f $(DESTDIR)$(PREFIX)/include/kmyth/kmyth.h
+	rm -f $(DESTDIR)$(PREFIX)/lib/$(TPM_UTIL_LIB_SONAME)
+	rm -f $(DESTDIR)$(PREFIX)/lib/$(LOGGER_LIB_SONAME)
+	rm -rf $(DESTDIR)$(PREFIX)/include/kmyth
```
PircDef, line 481: If someone has installed the sgx portion of the tool, this will unintentionally remove that as well. This should remove specific headers.

bit-bender: Good point ... I changed to delete the specific headers and then the directory, only if it is empty

bit-bender:
I think I fixed the problem with make test failing by adding the 'bin' directory as a order-only pre-requisite for the rule to build the kmyth-test executable (i.e., have make test create the bin directory if it wasn't created by the make tpm-util-lib)

tpm2/Makefile
```diff
@@ -351,100 +385,134 @@ doc: $(HEADER_FILES) \
...
 	indent $(INDENT_OPTS) $(TEST_SOURCES)
 	indent $(INDENT_OPTS) $(TEST_HEADERS)
-	rm -f test/src/*.c~
-	rm -f test/src/*/*.c~
-	rm -f test/include/*.h~
-	rm -f test/include/*/*.h~
-	mkdir -p $(TEST_OBJECT_DIRS)
 
-kmyth-test: $(TEST_OBJECTS)
+.PHONY: kmyth-test
```
b-carr, line 396: Rather than making the target phony would it be better to make the util lib a prerequisite for the test target?

tpm2/Makefile
```diff
@@ -254,91 +255,124 @@ INDENT_OPTS += -l80#                     max non-comment line length is 80
...
-kmyth-lib: $(LOGGER_OBJECTS) \
-           $(UTIL_OBJECTS) \
-           $(CIPHER_OBJECTS) \
-           $(TPM_OBJECTS)
+.PHONY: pre
+pre:
+	indent $(INDENT_OPTS) $(SOURCE_FILES)
+	indent $(INDENT_OPTS) $(HEADER_FILES)
+
+.PHONY: logger-lib
```
b-carr, line 268: Do this target really need to be phony?

tpm2/Makefile
```diff
@@ -254,91 +255,124 @@ INDENT_OPTS += -l80#                     max non-comment line length is 80
...
+logger-lib: pre  clean-backups $(LOGGER_OBJECTS) | $(LIB_DIR)
 	$(CC) $(SOFLAGS) \
 	      $(LOGGER_OBJECTS) \
+	      -o $(LOGGER_LIB_LOCAL_DEST) \
+	      $(LDLIBS)
+
+.PHONY: tpm-util-lib
```
b-carr, line 275: Instead of making the target phony can we add `logger-lib` as a prerequisite so it gets built anytime either one of it's sources changes or the logger changes?

tpm2/Makefile
bit-bender: This is now $(BIN_DIR)/kmyth-test instead of the phony kmyth-test target.

tpm2/Makefile
bit-bender: So, I made $(LIB_DIR)/libkmyth-logger.so a non-phony target, but I liked having a simpler make target name, so I kept a make logger-lib phony target that simply has $(LIB_DIR)/libkmyth-logger.so as a pre-requisite.

tpm2/Makefile
bit-bender: Similarly as above, I made $(LIB_DIR)/libkmyth-tpm.so a non-phony target but changed 'tpm-util-lib' to the more accurate and simpler 'libs' phony target, which has $(LIB_DIR)/libkmyth-logger.so and $(LIB_DIR)/libkmyth-tpm.so as pre-requisites.

Now, for *make all* or *make libs*, the libraries will get built based on changes to any pre-reqs. For *make logger-lib*, only the logger library will get built based on changes to its pre-reqs.

I think we want the 'logger-lib' to ignore the pre-reqs to the TPM utilities, because we don't want someone to have deal with TPM issues if all they want is the logger library.

tpm2/Makefile
bit-bender: I realized after looking at this again, I missed the part of this comment about adding the library as a prerequisite for the $(BIN_DIR)/kmyth-test target.

Now, if you call *make test* before building, it will build the code, test code, and run the tests.

In doing this, I also realize that if $(LIB_DIR)/libkmyth-logger.so is a prerequisite of $(LIB_DIR)/libkmyth-tpm.so, then by making the TPM utility library a pre-req, the logger library is indirectly a pre-req. Thus, I cleaned up a few places where I listed both libraries as pre-reqs.

bit-bender:
As part of this pull request, we moved making the build directories out of the phony 'pre' and 'pretest' targets and instead made non-phony directory targets with a 'mkdir -p' recipe and then made these directory targets order-only prerequisites to the appropriate higher-level targets so that the directories are created when needed.

For *make doc*, however, we still had the 'mkdir -p $(DOC_DIR)' as part of the doc target's recipe. Therefore, the last commit changes this to be consistent with how this is done elsewhere in the makefile.

PR #119, opened 2021-07-02
Add an implementation of Needham-Schroeder-Lowe for key retrieval

PeterHamilton:
This change adds a collection of functionality that implements the Needham-Schroeder-Lowe (NSL) protocol to support secure key retrieval without using TLS. NSL is used to create a shared session key between two parties which is then used to secure key requests and responses between the two parties.

NSL utility functions can be found in nsl_util.h/c, socket utility functions can be found in socket_util.h/c, and KMIP utility functions can be found in kmip_util.h/c. Exemplars of the two communicating parties can be found in the nsl_client.c and nsl_server.c applications. Both support CLI arguments; use -h for more information when running them.

PR #122, opened 2021-07-21
Moved new files added in PR #119 consistent with organization changes introduced by PR #120

bit-bender:


bit-bender:
When I installed the algorithm test vectors to run a 'make test' using them, they showed up as untracked files because, although in PR #120 we changed the path to the test directory, we forgot to update the .gitignore to reflect that change. The most recent commit for this pull request attempts to correct that issue. I also, generically added .pem files, as I created some to test the NSL code and it seems that we probably never want to inadvertently insert a .pem file into our software repo.

PR #124, opened 2021-08-05
Add an initial Diffie-Hellman implementation inside an SGX enclave

PeterHamilton:
This change updates the existing SGX enclave files adding support for Diffie-Hellman key exchange functionality. This initial update loads two private key pairs into the SGX enclave and computes the shared Diffie-Hellman secret and then derives a session key from that secret.

A test application, enclave configuration file, and Makefile are included to drive and test the SGX enclave functionality but are not intended for long term use.


PeterHamilton:
I didn't include any build instructions in the commit, so here are the highlights. Running the Makefile will produce an `app` executable in the same directory. It takes two CLI arguments (no usage or flags used right now) which are each file paths to unencrypted DER-encoded DH private key files. I can provide the OpenSSL CLI commands I used to generate them if needed. As mentioned in the `kmyth_enclave_ecalls.cpp` block comments, I've had issues getting OpenSSL to properly load the DH public key file, so for now I'm just extracting the public key from the private key file.

Running `app` with those two arguments should produce a couple text output detailing loading the key files and then, after the enclave code executes, two lines like this:

```
enc_run_dh_key_exchange return value: 0
Error buffer:
```

If the DH code ran successfully in the enclave the return value should be `0`. If it is not, an error occurred along the way; see the code for possible error values. In some places I copy in the OpenSSL error strings to help debug what's going wrong. That will be printed after `Error buffer:` if it's not empty.

b-carr:
We're going to have to reconcile this with #125, I think most of it should merge easily but ```kmyth_enclave_ecalls.cpp``` got renamed ```kmyth_enclave_trusted.cpp``` because (as I noticed from this pull request) there's lots that needs to be in there that isn't actually exposed through an ecall.

PeterHamilton:
@b-carr Agreed, I think this can fold in pretty seamlessly to the changes in #125. I debated about splitting my changes out into separate enclave files but it was easier to use what was already there. We can definitely reevaluate that as we continue to add functionality.

PircDef:
This pull request has been closed because it has been superseded by #142 #145 #149 and #157.

PR #125, opened 2021-08-05
Tests/sgx example and tests

b-carr:
This adds some basic tests of the kmyth SGX enclave ecalls, as well as providing an example (and documentation) on how to include the SGX functionality in another project.


PircDef:
There is an error because there is no signing key for the tests:

    Makefile:226: *** FAIL - No Enclave Signing Key found
    Generate or install sgx/kmyth_sgx_test_enclave_private.pem
    e.g., run 'openssl genrsa -out sgx/kmyth_sgx_test_enclave_private.pem -3 3072'.  Stop.

When someone is using the enclave with a real application, it's fair to require them to create this key. For the sake of these tests, I think we should generate a key, test it, then we can delete the .pem afterward.

PircDef (approved):

PR #126, opened 2021-08-06
Made Verbosity of Log Messages sent to User's Screen more Configurable

### LINKED ISSUE #123: Make logger prefix optional for logging to screen
While it contains useful information, the prefix that precedes each log entry can be distracting when monitoring normal activity via stddest (stdout / stderr). If it were more configurable, the user would have the option to choose between the current full information view or a streamlined (less cluttered) log entry view.

The initial idea is to always include the prefix when entries are logged to file. When logged to screen, however, if the logging severity level is LOG_INFO or higher, however, logging to screen would omit the prefix. If the logging severity level is lower (e.g., LOG_DEBUG), however, the prefix would be included. The user would then, for normal operation, be presented simplified log entries, but would be able to obtain the full log entries (prefix included) by selecting the verbose logging option.

bit-bender:
Log entries for the kmyth logger can contain information that is distracting when monitoring normal activity via stddest (stdout / stderr). As all of this information is potentially useful, the approach taken in this pull request is to make the logging verbosity slightly more configurable for the user.

The log message prefix currently contains the application name and version. The streamlined (less cluttered) logging view implemented by this pull request removes that information and shows only the severity level of the message (e.g., ERROR, WARNING, INFO, ...) as its prefix.

The log message contents include information about the source file location of the message. While this is important when debugging, it is probably just distracting when performing normal monitoring of an application. Therefore, this information is omitted from the streamlined view.

While not a change, I will point out that the timestamp for the message is currently only included in messages logged to file. It seems less useful for the more real-time monitoring performed by monitoring information via stdout/stderr. This pull request does not change this approach.

This pull request implements the following approach for determining which view to present the user (full or simplified) when logging to the screen:
- lf the logging severity level is LOG_INFO or less (not LOG_DEBUG), the simplified (less cluttered) message is displayed
- If the logging severity level is higher than LOG_INFO (e.g., LOG_DEBUG), however, the full log message would be displayed. 

In practice, then, he user would, for normal operation, be presented simplified log entries, but would be able to obtain the full log entries by selecting the --verbose (-v) command line option.

Closes #123.

### DIFF
--- a/logger/src/kmyth_log.c
+++ b/logger/src/kmyth_log.c
@@ -340,10 +340,10 @@ void log_event(const char *src_file,
       //
       //   for stddest:
       //     - timestamp is always omitted
-      //     - if logging severity threshold is <= LOG_INFO (<= 5), simplify
+      //     - if logging severity threshold is <= LOG_INFO (<= 6), simplify
       //       prefix to the severity level and content to the actual message
       //       (routine logging to a user's screen will be simplified).
-      //     - if logging severity threshold is > LOG_INFO (>5), include the
+      //     - if logging severity threshold is > LOG_INFO (> 6), include the
       //       full prefix (with application name/version) and enhance message
       //       with source location information. User can turn on detailed
       //       logging by using the --verbose (or -v) command line option.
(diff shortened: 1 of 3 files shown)


bit-bender:
I believe the changes to src/util/kmip_util.c and src/util/nsl_util.c are because of the indent utility. I did not manually edit these files.

bit-bender:
Per email discussion, this has been reviewed and approved, so I am going ahead with the merge.

PR #133, opened 2021-09-21
Allocating memory for an empty file now returns a NULL pointer.

jarednecaise:
malloc(0) is implementation dependent and may have unexpected results. Now returning a NULL pointer instead.

### DIFF
--- a/src/util/file_io.c
+++ b/src/util/file_io.c
@@ -164,8 +164,7 @@ int read_bytes_from_file(char *input_path, uint8_t ** data,
 
   if (input_size == 0)
   {
-    kmyth_log(LOG_DEBUG, "input file size is zero, returning NULL ... exiting");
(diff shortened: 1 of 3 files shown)


src/util/file_io.c
```diff
@@ -162,6 +162,18 @@ int read_bytes_from_file(char *input_path, uint8_t ** data,
   }
   int input_size = st.st_size;
 
+  if (input_size == 0)
+  {
+    kmyth_log(LOG_DEBUG, "input file size is zero, returning NULL ... exiting");
```
PircDef, line 167: It isn't necessarily expected to exit because we aren't returning an error, but I probably wouldn't kmyth_log any debug message for this.  Also, your if(!BIO_free) has an extra tab.

b-carr (approved):

PR #134, opened 2021-09-23
Build kmyth general utilities as a shared library

bit-bender:
In order to provide applications utilizing kmyth more flexible access to underlying kmyth utilities, without copy/paste or some other means of including kmyth utility source files, this pull request restructures kmyth to build a 'kmyth-utils' shared library that can be linked in to expose this kmyth utility functionality.

In order to hopefully make this cleaner, the "file_io", "formatting_tools", and "memory_util" sources were moved into a top-level "utils" directory and used to build a "libkmyth-utils.so" library (in the lib folder). The remaining files in "src/util" and "include/util" were put into either "src/network", "src/protocol", "include/network", or "include/protocol".

An attempt was made to change Makefile variable names not related to the "kmyth-utils" library having a "util" or "utility" reference to something different or less confusing.

A "utils-lib" target (make utils-lib) was added to build only the new library. The "libs", "docs", "test", "install", and "uninstall" targets were updated to reflect the changes.

The build information in the INSTALL.md file was updated.


b-carr (commented):
It looks like this only changes the build, with a few minor changes resulting to the code, or did I miss something?

bit-bender:
Yes, you are correct. This builds the code in formatting_tools.c, file_io.c, and memory_util.c as a shared library. The rest of the kmyth code (minus the logger code), then, is built by linking against this library. This provides an API with access to these "kmyth utilities" (e.g., base64 encode/decode, ...) by other applications.

In an attempt to avoid confusion, this pull request also reorganizes and renames a little to avoid re-use of the term "utilities".

The changes to the code are indeed minor, but I think the "indent" utility caused some fairly widespread reformatting of files.

b-carr (approved):

PR #140, opened 2021-09-29
Proposed SGX Code Restructure

bit-bender:
This pull request proposes a fundamental restructure of the kmyth SGX repository. Key attributes of this restructure are:
- divide 'kmyth_enclave' code into 'untrusted' (OCALL and other untrusted code that can be "inherited" by an application", 'trusted' (ECALL and other trusted code that can be "inherited by an application), and 'common' (functions that can be invoked from either trusted or untrusted space)
- adds a 'demo' directory
- divides 'test' and 'demo' code into 'app' and 'enclave' directories

Edits to the 'test' folder were only made to adapt the build for the existing test application.

The 'demo' application added represents an almost empty "retrieve key from key server" application. It reads from file and converts to DER format, both a key and certificate. After initializing an SGX enclave, it uses the "kmyth_enclave_retrieve_key_from_server()" ECALL to pass these into to enclave, where it unmarshals them. The intent was to include simple code examples from all kmyth_enclave code categories (untrusted, trusted, and common).


sgx/common/include/ec_key_cert_marshal.h
```diff
@@ -0,0 +1,74 @@
...
+ *                                   a DER formatted byte array)
+ *
+ * @param[out] ec_der_bytes_out      Pointer to byte array that contains
+ *                                   the marshalled EC private key result
+ *
+ * @param[out] ec_der_bytes_out_len  Pointer to size of the byte array
+ *                                   containing the marshalled EC private
+ *                                   key result
+ */
+int marshal_ec_pkey_to_der(EVP_PKEY ** ec_pkey_in,
```
b-carr, line 47: What do the return values of these functions mean?

sgx/common/include/ec_key_cert_unmarshal.h
```diff
@@ -0,0 +1,61 @@
...
+ *                                 key input
+ *
+ * @param[in] ec_der_bytes_in_len  Pointer to size of the byte array
+ *                                 containing the marshalled (DER format)
+ *                                 EC private key input
+ *
+ * @param[out] ec_pkey_out         Pointer to a EVP_PKEY struct to hold the
+ *                                 unmarshalled EC key result
+ */
+int unmarshal_ec_der_to_pkey(uint8_t ** ec_der_bytes_in,
```
b-carr, line 36: What do the return values mean?

sgx/common/src/ec_key_cert_marshal.c
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  // Validate that a pointer to a NULL buffer pointer was passed in for
+  // the binary output byte array. Put it in that state if necessary.
+  // This will support memory allocation by the call to i2d_PrivateKey()
+  // for PKEY to DER format conversion rather than having to do the
+  // required size calculation and memory allocation beforehand.
+  if (*ec_der_bytes_out != NULL)
+  {
+    free(*ec_der_bytes_out);
```
b-carr, line 40: I'd prefer to require that ```*ec_der_bytes_out``` be ```NULL``` and have the function error if it's not. As written this could segfault if the caller passes an uninitialized pointer.

sgx/common/src/ec_key_cert_marshal.c
```diff
@@ -0,0 +1,112 @@
...
+    *ec_der_bytes_out = NULL;
+  }
+
+  // A copy of the pointer is needed because i2d_PrivateKey() will
+  // modify the one it is passed as a parameter.
```
b-carr, line 45: Just for my clarity, it modifies the pointer but not the pointed-to-data?

sgx/common/src/ec_key_cert_marshal.c
```diff
@@ -0,0 +1,112 @@
...
+int marshal_ec_x509_to_der(X509 ** ec_cert_in,
+                           unsigned char **ec_der_bytes_out,
+                           int *ec_der_bytes_out_len)
+{
+  // Validate that a pointer to a NULL buffer pointer was passed in for
+  // the binary output byte array. Put it in that state if necessary.
+  // This will support memory allocation by the call to i2d_X509()
+  // for X509 to DER format conversion rather than having to do the
+  // required size calculation and memory allocation beforehand.
+  if (*ec_der_bytes_out != NULL)
```
b-carr, line 82: Same thing here, I'd rather error out of it's not ```NULL```.

sgx/common/src/ec_key_cert_unmarshal.c
```diff
@@ -0,0 +1,46 @@
...
+ * @brief Provides implementation for converting binary (DER) formatted
+ *        keys and/or certificates into structs used by the OpenSSL API.
+ */
+
+#include "ec_key_cert_unmarshal.h"
+
+/*****************************************************************************
+ * unmarshal_ec_der_to_pkey()
+ ****************************************************************************/
+int unmarshal_ec_der_to_pkey(uint8_t ** ec_der_bytes_in,
```
b-carr, line 13: What happens if ```*ec_der_bytes_in``` is ```NULL```? Does ```d2i_PrivateKey``` then return ```NULL``` so we catch it in that call?

sgx/common/src/ec_key_cert_unmarshal.c
```diff
@@ -0,0 +1,46 @@
...
+    return EXIT_FAILURE;
+  }
+
+  return EXIT_SUCCESS;
+}
+
+/*****************************************************************************
+ * unmarshal_ec_der_to_x509()
+ ****************************************************************************/
+int unmarshal_ec_der_to_x509(uint8_t ** ec_der_bytes_in,
```
b-carr, line 32: What happens if ```*ec_der_bytes_in``` is ```NULL```?

sgx/demo/enclave/kmyth_sgx_retrieve_key_demo_enclave.edl
```diff
@@ -0,0 +1,6 @@
+enclave
+{
+  from "kmyth_enclave.edl" import *;
+	from "sgx_tsgxssl.edl" import *;
```
b-carr, line 4: GitHub is showing me an odd spacing here, not sure if it's real or not.

sgx/trusted/include/kmyth_enclave.h
```diff
@@ -0,0 +1,46 @@
+#ifndef _KMYTH_ENCLAVE_H_
```
b-carr, line 1: Should we rename this file something like ```kmyth_enclave_trusted_h``` to indicate in the filename that it's trusted functionality (and shouldn't be needed in untrusted code)?

sgx/common/include/ec_key_cert_marshal.h
bit-bender: Added comment: @return 0 on success, 1 on error

sgx/common/include/ec_key_cert_unmarshal.h
bit-bender: Added comment: @return 0 on success, 1 on error

sgx/common/src/ec_key_cert_marshal.c
bit-bender: Yes, I originally did it that way and then thought that I could make it work even if the caller passes in an already allocated pointer. If an uninitialized pointer is passed in, though, this is still problematic. I replaced the free and assign to NULL statements with return failure. I removed the log message because when I added log messages to the unmarshal code (currently used by trusted code), I realized that even though log_event_ocall() is is known within trusted and untrusted code, the include path is dependent upon where you call it.

sgx/common/src/ec_key_cert_marshal.c
bit-bender: For the byte array, I believe that it increments the pointer you pass in and returns the incremented pointer (e.g., a value pointing to the end of the array). I was getting a lot of seg-faults when using these pointers later and I think that this was the reason. I do not know why I also made a copy of the pointers to the EVP_PKEY and X509 structs, so I just changed the code in these two marshal functions to use the passed in pointers to those structs directly.

Below are some excerpts from OpenSSL documentation related to this:

The ways that *ppin and *ppout are incremented after the operation can trap the unwary. See the WARNINGS section for some common errors. The reason for this-auto increment behaviour is to reflect a typical usage of ASN1 functions: after one structure is encoded or decoded another will be processed after it.

...

WARNINGS

Using a temporary variable is mandatory. A common mistake is to attempt to use a buffer directly as follows:

 int len;
 unsigned char *buf;

 len = i2d_X509(x, NULL);
 buf = OPENSSL_malloc(len);
 ...
 i2d_X509(x, &buf);
 ...
 OPENSSL_free(buf);
This code will result in buf apparently containing garbage because it was incremented after the call to point after the data just written. Also buf will no longer contain the pointer allocated by OPENSSL_malloc() and the subsequent call to OPENSSL_free() is likely to crash.

...

sgx/common/src/ec_key_cert_marshal.c
bit-bender: Same as above. I replaced the free and assign to NULL statements with return failure

sgx/common/src/ec_key_cert_unmarshal.c
bit-bender: I don't think d2i_PrivateKey() checks for a NULL pointer being passed in. I didn't look at the OpenSSL source code, but when I tried it (substituted the &buf_in parameter with NULL) it core dumped. I added a check for this.

sgx/common/src/ec_key_cert_unmarshal.c
bit-bender: Similar to above, I added a check to make sure the passed in parameter is non-NULL.

sgx/trusted/include/kmyth_enclave.h
bit-bender: This is a good suggestion. I renamed the file to kmyth_enclave_trusted.h.

sgx/demo/enclave/kmyth_sgx_retrieve_key_demo_enclave.edl
bit-bender: I think I fixed this by editing in GitHub

sgx/README.md
```diff
@@ -31,31 +35,30 @@ SGX_SSL_INCLUDE_PATH ?= <path to SGX SSL headers>
 ```
 * The location of the ```kmyth_enclave.edl``` file is included in the search path for ```SGX_EDGER8R``` along with ```$(SGX_SSL_INCLUDE_PATH)```
 ```
-sgx/kmyth_sgx_test_enclave_t.c: $(SGX_EDGER8R) sgx/kmyth_sgx_test_enclave.edl
-	@cd sgx && $(SGX_EDGER8R) --trusted kmyth_sgx_test_enclave.edl --search-path $(SGX_SDK)/include --search-path . --search-path ../../../sgx/kmyth_enclave --search-path $(SGX_SSL_INCLUDE_PATH)
+enclave/kmyth_sgx_test_enclave_t.c: $(SGX_EDGER8R) enclave/kmyth_sgx_test_enclave.edl
```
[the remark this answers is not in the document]
bit-bender: I agree. I have not yet updated this, but will. I will follow-up here after I have done so.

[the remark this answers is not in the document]
bit-bender: I took a shot at updating the README to address this.

sgx/common/src/ec_key_cert_marshal.c
bit-bender: I mentioned in the previous comment that implementing logging in the "common" functions was problematic because for the trusted code, the declaration of log_ocall() is via the EDGR8R generated header file and for the untrusted code via the log_ocall.h header file directly. In an effort to address this, I renamed the "enclave_log()" macro to "kmyth_sgx_log()" and moved it to a new kmyth_enclave_common.h header file. I added a definition of _KMYTH_LOCALE_TRUSTED_ kmyth_enclave_trusted.h. and added conditional includes in kmyth_enclave_common.h based on the value of this variable. My thought was that if all  "common" functionality is included by kmyth_enclave_trusted.h including kmyth_enclave_common.h, that this would be fairly clean, but this is still a little messy and may or may not be the right approach in the big picture. I also added a temporary test block in the demo app to show that using kmyth_sgx_log() from the untrusted side works.

b-carr (commented):
I'm happy with these, just double check that you've got all the function return values documented.

bit-bender:
I think I checked and fixed the missing return value documentation. Somehow the common/include/kmyth_enclave_common.h header file was missing from this branch. I fixed this and verified that the tests and demo build and run correctly again.

PircDef (approved):

PR #142, opened 2021-10-20
Initial kmyth enclave 'retrieve key' ECDH key agreement implementation

bit-bender:
Add initial kmyth enclave 'retrieve key' functionality:
- adds an ECALL to invoke from untrusted space
- ECALL calls a 'wrapper' function that provides an API call from within the enclave
- Currently only implements an ECDH key agreement between the enclave and a remote peer
    - OCALL facilitates exchange of ephemeral 'public key' contributions between enclave and remote peer
    - OCALL currently calls simply dummy_ecdh_server() function - to be replaced by code to integrate with a real server
 - Attempted to modularize with functionality in common/src/ecdh_util.c


sgx/common/src/ecdh_util.c
```diff
@@ -0,0 +1,343 @@
...
+int create_ecdh_ephemeral_key_pair(int ec_nid,
+                                   EC_KEY ** ephemeral_ec_key_pair_out)
+{
+  // create new EC_KEY object for the specified built-in curve
+  //   The EC_KEY object passed to 'generate_key' below must be associated
+  //   with the desired EC_GROUP.
+  *ephemeral_ec_key_pair_out = EC_KEY_new_by_curve_name(ec_nid);
+  if (*ephemeral_ec_key_pair_out == NULL)
+  {
+    kmyth_sgx_log(3, "failed to create new elliptic curve key object by NID");
```
PircDef, line 23: Is there a reason not to include syslog.h for the log severity (e.g., LOG_ERR)?

bit-bender: syslog.h did not seem to be available inside the enclave, so I moved on when I was working on this. I will look into this further as I agree the magic numbers for things like log message severity levels is suboptimal.

bit-bender:
I added a line to the test key/cert generation bash script to write a ~/.rnd file before generating test keys/certs. This fixes the "Can't load ~/.rnd into RNG" warnings that were previously generated.

I fixed the Makefile so that 'make' now is equivalent to 'make all'. I revised the top-level build targets as follows:
- **pre**: creates the ./bin directory, generates the enclave signing key, runs the 'indent' utility on source files, and cleans up automatic backup files.
- **all**: builds the ./bin/kmyth_sgx_retrieve_key_demo binary
- **test-keys-certs**: generates the .pem test key and certificate files in the ./data director
- **demo**: builds the code (all target), generates the test keys/certs (test-keys-certs target), and runs the demo binary

bit-bender:
In the original submission, most "length" variables were of type int (signed integer). The OpenSSL API, however, appropriately expects either size_t (unsigned long int) or unsigned int. The latest commit implements variable type changes to make this code more consistent.

When doing this, I also came across a redundant definition of the "dummy_server" function in sgx/common/include/ecdh_util.h that I removed and two wrongly named variables in the compute_ecdh_shared_secret() function definition that I corrected.

JWZeta:
Needs to be merged with main to Incorporated all of the latest merges.

sgx/common/src/ecdh_util.c
bit-bender: In the most recent commit (9045fff), I think I addressed the use of the logging priority macros defined in syslog.h. For now, if they are not already defined, I define these macros in sgx/common/include/kmyth_enclave_common.h. Longer term, there is probably a more elegant way to include them directly from syslog.h inside the enclave.

PircDef (approved):

sgx/common/src/ecdh_util.c
```diff
@@ -0,0 +1,343 @@
...
+ * compute_ecdh_shared_secret()
+ ****************************************************************************/
+int compute_ecdh_shared_secret(EC_KEY * local_eph_priv_key,
+                               EC_POINT * remote_eph_pub_point,
+                               unsigned char ** shared_secret,
+                               int * shared_secret_len)
+{
+  // create buffer (allocate memory) for the shared secret (session key) result
+  int field_size = EC_GROUP_get_degree(EC_KEY_get0_group(local_eph_priv_key));
+  *shared_secret_len = (field_size + 7) / 8;
```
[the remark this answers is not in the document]
bit-bender: I added an explanation for this.

sgx/trusted/include/wrapper/sgx_retrieve_key_impl.h
```diff
@@ -0,0 +1,57 @@
...
+#include <openssl/evp.h>
+#include <openssl/err.h>
+
+#include "kmyth_enclave_trusted.h"
+
+/**
+ * @brief Retrieve a designated key from a "remote" key server securely
+ *        into the enclave.
+ *
+ *        TODO: The parameters to this function will have to be augmented
```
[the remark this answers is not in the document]
bit-bender: I went ahead and addressed the two TODO items in the code. The first (parameter expansion of the 'retrieve_key' functionality) is a natural outcome of the planned evolution of this code. For this reason, I deleted this TODO without creating an issue. I created issue #146 for the second, investigation of specifying a KDF in the compute_ECDH_key() call to simplify code.

PR #145, opened 2021-11-10
Add test ECDH key server

arp102:
~~These changes are on top of open PR #142~~

Build/usage/implementation details are in sgx/demo/server/README.md

The source is located in sgx/demo because it reuses
code from sgx/common/src/ecdh_util.c
It also includes header files from outside the sgx directory
that are required to import from libkmyth-tpm.

### DIFF
--- a/sgx/demo/server/ecdh_server.c
+++ b/sgx/demo/server/ecdh_server.c
@@ -43,9 +43,9 @@ typedef struct ECDHServer {
   EVP_PKEY *remote_pubkey;
   EC_KEY *local_ephemeral_keypair;
   unsigned char *remote_ephemeral_pubkey;
-  int remote_ephemeral_pubkey_len;
+  size_t remote_ephemeral_pubkey_len;
   unsigned char *session_key;
-  int session_key_len;
+  unsigned int session_key_len;
   struct sockaddr_storage peer_addr;
   socklen_t peer_addr_len;
 } ECDHServer;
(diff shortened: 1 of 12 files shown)


sgx/demo/server/ecdh_server.c
```diff
@@ -0,0 +1,580 @@
...
+    kmyth_log(LOG_ERR, "signature of ECDH remote 'public key' invalid");
+    error(this);
+  }
+  kmyth_log(LOG_DEBUG, "validated signature on ECDH remote 'public key'");
+}
+
+void send_ephemeral_public(ECDHServer *this)
+{
+  unsigned char *local_pub, *local_pub_sig;
+  int local_pub_len, local_pub_sig_len;
```
PircDef, line 391: local_pub_len should be size_t

sgx/demo/server/ecdh_server.c
```diff
@@ -0,0 +1,580 @@
...
+
+  kmyth_clear_and_free(local_pub, local_pub_len);
+  kmyth_clear_and_free(local_pub_sig, local_pub_sig_len);
+}
+
+void get_session_key(ECDHServer *this)
+{
+  EC_POINT *remote_ephemeral_pub_pt = NULL;
+  unsigned char *session_secret = NULL;
+  int session_secret_len = 0;
```
PircDef, line 429: session_secret_len should be size_t

arp102:
The ECDH and NSL test applications now use TCP instead of UDP.
(It was easier to do both because of the shared util code in libkmyth.)

The ECDH application now has separate source files and binaries for the client and server components.
(The bulk of their code is still in a shared source file.)
If necessary, I can further reorganize the ECDH code into src/include directories.

PircDef (approved):

PR #149, opened 2022-01-04
Full key exchange in SGX demo

arp102:
Working ECDHE SGX demo.

This requires the modified libkmip in https://github.com/arp102/libkmip/tree/dev-kmyth
Still need to determine how that change should be incorporated into the libkmip master branch.

After reinstalling libkmip, use the `make clean demo` command to automatically execute both demo programs.

Should probably clean up some of the log output to make it more readable.

**Resolved:** Still need to clean up the code in sgx_retrieve_key_impl.c to enable or remove the commented log messages (unless we deduplicate all of those copied kmyth functions).

**Resolved:** Some functions in sgx_retrieve_key_impl.c are copied from other kmyth sources. ~~but it probably wouldn't be practical to deduplicate them.~~
We will deduplicate the code from aes_gcm.c (and possibly from kmip_util.c), which requires some refactoring for compatibility with SGX.


arp102:
Rebased to incorporate PR #150, moved the socket initialization, and deduplicated the aes_gcm functions.

arp102:
FYI, the current demo uses a hard-coded timestamp in the kmip request because getting the real timestamp would require a new ocall.
Would that be worth adding?

arp102:
I’ve also looked at deduplicating the functions that I copied from `kmip_util.c` to `sgx_retrieve_key_impl.c`, but that will require some more complicated refactoring than `aes_gcm.c` did.
I wasn’t sure if you’d want me to make that change as well or if you’d think it’s more trouble than it’s worth.

The changes would probably look like:
1. Move the `retrieve_key_with_session_key` and `send_key_with_session_key` functions into a separate source file because they do network IO.
2. Replace the incompatible `kmyth_log` and `time` calls using the preprocessor when building the enclave version of `kmip_util`.

sgx/trusted/src/wrapper/sgx_retrieve_key_impl.c
```diff
@@ -212,9 +412,100 @@ int enclave_retrieve_key(EVP_PKEY * enclave_sign_privkey, X509 * peer_cert)
...
+
+  snprintf(msg, MAX_LOG_MSG_LEN, "Received a KMIP object with ID: %.*s",
+           (int) received_key_id_len, received_key_id);
+  kmyth_sgx_log(LOG_DEBUG, msg);
+
+  snprintf(msg, MAX_LOG_MSG_LEN, "Received operational key: 0x%02X..%02X",
+           key[0], key[key_len - 1]);
+  kmyth_sgx_log(LOG_DEBUG, msg);
+
   kmyth_sgx_log(LOG_DEBUG, "completed ECDH exchange");
```
bit-bender, line 509: Now that the demo is more than the ECDH key agreement, this log message should probably be updated to something like "completed key retrieval from server into enclave"

sgx/untrusted/src/ocall/ecdh_ocall.c
```diff
@@ -15,33 +50,142 @@ int ecdh_exchange_ocall(unsigned char *enclave_ephemeral_public,
...
+                        unsigned int *remote_eph_pub_signature_len,
+                        int socket_fd)
+{
+  int ret;
+
+  kmyth_log(LOG_DEBUG, "Sending ephemeral public key.");
+  ret =
+    write(socket_fd, &enclave_ephemeral_public_len,
+          sizeof(enclave_ephemeral_public_len));
+  if (ret == -1)
```
bit-bender, line 64: This is probably kind of a trivial comment, but is it more robust to check if ret != 0 instead of checking that ret == -1. I actually am not an expert on the socket API, so I am good with whatever is considered best practice. Note: this applies to a number of conditionals in this file, but I will only mention it here.

sgx/untrusted/src/util/ecdh_dummy_server.c
```diff
@@ -16,7 +16,7 @@ int ecdh_dummy_server(unsigned char *client_pub,
                       unsigned char *client_pub_sig,
```
bit-bender, line 16: Is ecdh_dummy_server.c (and ecdh_dummy_server.h) unneeded at this point? If so, I think we should probably go ahead and delete them.

bit-bender:
I think that we should either add the OCALL to get the real timestamp to this pull request or leave it hard-coded for now, but add it as an issue to be resolved at some later point.

sgx/trusted/src/wrapper/sgx_retrieve_key_impl.c
```diff
@@ -6,11 +6,203 @@
...
+
 //############################################################################
 // enclave_retrieve_key()
 //############################################################################
-int enclave_retrieve_key(EVP_PKEY * enclave_sign_privkey, X509 * peer_cert)
+int enclave_retrieve_key(EVP_PKEY * enclave_sign_privkey, X509 * peer_cert,
+                         const char *server_host, int server_host_len,
+                         const char *server_port, int server_port_len)
 {
+  int ret_val, ret_ocall;
```
bit-bender, line 193: The variable ret_ocall should be of type sgx_status_t.

sgx/demo/Makefile
```diff
@@ -330,31 +350,27 @@ $(App_Name): $(App_Source_Files) \
              enclave/ecdh_dummy_server.o \
```
bit-bender, line 350: This is related to an earlier comment, but if we do get rid of the "dummy_server" code, we will need to clean up the Makefile also.

bit-bender:
I just realized that I never addressed your question re: deduplicating the functions that I copied from kmip_util.c to sgx_retrieve_key_impl.c.

I do believe that this is worthwhile. I am not convinced, however, that this refactoring needs to necessarily be completed as part of this pull request. I think that creating an issue and addressing this in a follow-on pull request would also be acceptable.

Do you think that a more generic approach to using the preprocessor to replace functions that are incompatible inside the enclave is potentially a method we could have in our toolbox when we may encounter these types of issues elsewhere or does this have to be implemented in a very case-specific manner?

sgx/demo/Makefile
```diff
@@ -247,7 +259,7 @@ all: $(App_Name) $(Enclave_Lib)
 	@echo "To build the project in simulation mode set SGX_MODE=SIM. To build"
 	@echo "the project in prerelease mode set SGX_PRERELEASE=1 and SGX_MODE=HW."
 else
-all: pre $(App_Name) enclave/$(Signed_Enclave_Name)
+all: pre enclave/$(Signed_Enclave_Name) $(App_Name) $(Server_Name) $(Client_Name)
 endif
 
 pre:
```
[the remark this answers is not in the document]
arp102: Yes, I'll add those.

sgx/trusted/src/wrapper/sgx_retrieve_key_impl.c
arp102: Yes, that's better.

sgx/untrusted/src/ocall/ecdh_ocall.c
arp102: Yes, it's generally better to compare the return value to the expected read/write length.
This catches under-reads and under-writes (which aren't really a concern here)
and also detects when we read from a socket that just closed.

sgx/untrusted/src/util/ecdh_dummy_server.c
arp102: Yes, I can remove them.

sgx/trusted/src/wrapper/sgx_retrieve_key_impl.c
arp102: Okay, I'll fix that.

arp102:
I added the new time ocall, and the kmip messages use valid timestamps now.

For the kmip_util deduplication, I added a `KMYTH_SGX` variable to use for all conditional preprocessing.
It's being used now in `kmip_util.c` and in `defines.h`.

I think this kind of preprocessing would be a good approach for future migrations to SGX.
I made the change in `defines.h` to show it could be used generically without breaking anything.

arp102:
The duplicate enclave build commands in `sgx/demo/Makefile` and `sgx/test/Makefile` have been annoying to keep up to date.
I'd recommend merging them to avoid that trouble in the future.

bit-bender (approved):
I agree with the suggestion made as part of this pull request to look into merging largely duplicated build commands in sgx/test/Makefile and sgx/demo/Makefile. Rather then add even more to this already extensive pull request, I have approved this pull request without addressing that indirect issue but created issue #151 to capture this as a future improvement.

Issue #153, opened 2022-01-24
Update enclave_retrieve_key() API


bit-bender:
The current API for calling the 'retrieve-key' functionality from within the enclave needs modification to support its use beyond the current demo. The following initial issues have been identified:
- The 'key ID' parameter passed to the server is currently generated within the function. This needs to be exposed to the API to permit the caller of the enclave_retrieve_key() to specify this value
- The returned key object is not exposed to the caller. The API needs to provide a reference to this value to the caller so that it has access to it.
- The port value is specified as a string, should pass this as an integer rather than doing the string conversion as part of the function.

PR #157, opened 2022-01-31
Update enclave_retrieve_key() API

bit-bender:
This pull request addresses Issue #153.

A pointer to the retrieved key (and its length) were added to the parameter list in order to provide the caller (within the enclave) access to the retrieved key.

Key ID string (and length) parameters were added to support caller specification of the key to be retrieved from the server.
- for enclave_retrieve_key() the input value (used to format the 'retrieve key' request) and the output value (parsed from the 'retrieve key' response were passed as separate parameters.
- for the kmyth_enclave_retrieve_key_from_server() ECALL, only the input requested 'key ID' was added as a parameter. After its call to enclave_retrieve_key() returns, however, the ECALL code checks that the output (key ID parsed from the response) matches the input key ID requested and errors if the two mismatch.

The server's TCP port value parameter was changed from a string to an integer. In the setup_socket_ocall() function, a "service" parameter string, which is a string version of the port value is required, so integer to string conversion is added in this one place.

The parameter description documentation was missing for the kmyth_enclave_retrieve_key_from_server() ECALL, so this pull request added some to the .edl file.

### DIFF
--- a/sgx/trusted/src/ecall/kmyth_enclave_retrieve_key.cpp
+++ b/sgx/trusted/src/ecall/kmyth_enclave_retrieve_key.cpp
@@ -75,8 +75,8 @@ int kmyth_enclave_retrieve_key_from_server(uint8_t * client_private_bytes,
 
   char msg[MAX_LOG_MSG_LEN] = { 0 };
 
-  snprintf(msg, MAX_LOG_MSG_LEN, "Retrieved into enclave key with ID: %.*s",
-           (int) retrieve_key_result_id_len, retrieve_key_result_id);
+  snprintf(msg, MAX_LOG_MSG_LEN, "Retrieved into enclave key with ID: %s",
+           retrieve_key_result_id);
   kmyth_sgx_log(LOG_DEBUG, msg);
 
   snprintf(msg, MAX_LOG_MSG_LEN,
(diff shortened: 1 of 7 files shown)


sgx/trusted/src/ecall/kmyth_enclave_retrieve_key.cpp
```diff
@@ -55,19 +56,48 @@ int kmyth_enclave_retrieve_key_from_server(uint8_t * client_private_bytes,
...
   {
     kmyth_sgx_log(LOG_ERR,
                   "enclave_retrieve_key() wrapper function call failed");
     return EXIT_FAILURE;
   }
 
+  char msg[MAX_LOG_MSG_LEN] = { 0 };
+
+  snprintf(msg, MAX_LOG_MSG_LEN, "Retrieved into enclave key with ID: %.*s",
+           (int) retrieve_key_result_id_len, retrieve_key_result_id);
```
PircDef, line 79: This int cast should be checked because the id_len is a size_t

bit-bender: Yes, retrieve_key_result_id_len is of type size_t (unsigned int) and is cast to int because the '.*' precision value in the format string expects an int. I'm not sure what to check, because the uint -> int should never be negative result. If it was, however, the precision would be ignored, which would negate the intended protection against a potentially unterminated string.

The way I tried to fix this was to check that the string parsed out of the KMIP response indeed has the expected length in the enclave_retrieve_key() function and then removed the precision (and hence the need for the cast) from the format string in both that function and here since we have verified with the length check that the string is terminated.

bit-bender: I added a comment above the string length checks noting the restriction of this implementation handling the Key ID as a string and not supporting '\0' as a valid character within the Key ID value itself. As the documentation in the header and .edl files defines 'Key ID' parameters as string values, my thought was that they did not all need to include this further clarification.

PircDef (approved):

Issue #151, opened 2022-01-15
Investigate Merging Makefiles for SGX demo and test


bit-bender:
The duplicate enclave build commands in sgx/demo/Makefile and sgx/test/Makefile are annoying to keep up to date.
This issue recommends potentially merging them to avoid the issues with maintaining duplicated build specifications.

PircDef:
Closed with #158

PR #158, opened 2022-02-06
Improvement/merge makefiles

b-carr:
This merge addresses issue #151


sgx/Makefile
```diff
@@ -0,0 +1,687 @@
...
+define err_no_test_enclave_signing_key
+FAIL - No Test Enclave Signing Key found
+Generate or install test/enclave/$(Test_Enclave_Signing_Key)
+e.g., run 'openssl genrsa -out test/enclave/$(Test_Enclave_Signing_Key) -3 3072'
+endef
+
+.PHONY: test-pre test-all test-run demo-pre demo-all demo-test-keys-certs demo
+test-pre:
+	@mkdir -p test/bin 
+	@if [ ! -f test/enclave/$(Test_Enclave_Signing_Key) ]; then \
```
PircDef, line 295: Is there a compelling reason not to use the same singing key for both of the test enclaves?

b-carr: I don't think so, so 2ed25a1 changes it to use the same key for both.

PR #161, opened 2022-02-22
TLS proxy program for SGX demo

arp102:
## Changes

* Add man-in-the-middle proxy program `tls-proxy`, which accepts ECDH connections using our custom protocol so that SGX demo applications can communicate with standard TLS services.
* Change message protocol for ECDH connection, adding an unencrypted header containing the message length.
* Fix KMIP key_id bugs.
* Refactor some ecdh_demo code and remove kmip_io_util.


arp102:
## PyKMIP Usage

Warning 1: PyKMIP's certificate generation script `bin/create_certificates.py` does not include a Subject Alternative Name (SAN) in the server certificate, which is required for hostname verification (currently always enabled in the proxy). The SAN in the server certificate must match the proxy's `--remote-ip` argument. This fork of PyKMIP sets the SAN to `localhost` https://github.com/arp102/PyKMIP/tree/improvement/cert-san.

Warning 2: PyKMIP requires clients to use certificates signed by a known CA. The certificate generation script creates multiple client certificates that can be used by the proxy.

Warning 3: PyKMIP's example configuration files don't quite work out of the box. Working configurations are included here.

### Instructions

``` bash
#!/bin/bash

# Install PyKMIP
# This may require installing some apt/dnf dependencies as well.
pip3 install --user pykmip

# Create PyKMIP directories
sudo mkdir -p /var/log/pykmip /etc/pykmip/certs /etc/pykmip/policies
sudo chown -R "${USER}" /var/log/pykmip /etc/pykmip

# Generate certificates
cd /etc/pykmip/certs
curl -O https://raw.githubusercontent.com/arp102/PyKMIP/improvement/cert-san/bin/create_certificates.py
python3 create_certificates.py

# Create server config file
echo "
[server]
hostname=localhost
port=5696
ca_path=/etc/pykmip/certs/root_certificate.pem
key_path=/etc/pykmip/certs/server_key.pem
certificate_path=/etc/pykmip/certs/server_certificate.pem
auth_suite=TLS1.2
policy_path=/etc/pykmip/policies
enable_tls_client_auth=True
tls_cipher_suites=
    TLS_RSA_WITH_AES_128_CBC_SHA256
    TLS_RSA_WITH_AES_256_CBC_SHA256
    TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384
    ECDHE-ECDSA-AES256-GCM-SHA384
    ECDHE-RSA-AES256-GCM-SHA384
    ECDHE-ECDSA-AES256-SHA384
    ECDHE-RSA-AES256-SHA384
logging_level=DEBUG
database_path=/etc/pykmip/pykmip.db
" > /etc/pykmip/server.conf

# Create client config file
echo "
[client]
host=localhost
port=5696
keyfile=/etc/pykmip/certs/client_key_john_doe.pem
certfile=/etc/pykmip/certs/client_certificate_john_doe.pem
cert_reqs=CERT_REQUIRED
ssl_version=PROTOCOL_SSLv23
ca_certs=/etc/pykmip/certs/root_certificate.pem
do_handshake_on_connect=True
suppress_ragged_eofs=True
username=example_username
password=example_password
" > /etc/pykmip/pykmip.conf

# Run the server [in a separate terminal]
pykmip-server

# Register keys with the server
python3 -c "
from kmip import enums
from kmip.pie import client
from kmip.pie import objects

c = client.ProxyKmipClient()

symmetric_key = objects.SymmetricKey(
     enums.CryptographicAlgorithm.AES,
     128,
     (
         b'\x00\x01\x02\x03\x04\x05\x06\x07'
         b'\x08\x09\x0A\x0B\x0C\x0D\x0E\x0F'
     )
 )

with c:
    for i in range(10):
        key_id = c.register(symmetric_key)
        print(f'Registered new key with ID {key_id}')
"

# Build the kmyth programs
cd kmyth/sgx
make clean demo-all demo-test-keys-certs

# Run the proxy program using the pykmip CA and client cert [in a separate terminal]
./demo/bin/tls-proxy -r demo/data/server_priv_test.pem -u demo/data/client_cert_test.pem -p 7000 -I localhost -P 5696 -C /etc/pykmip/certs/root_certificate.pem -R /etc/pykmip/certs/client_key_john_doe.pem -U /etc/pykmip/certs/client_certificate_john_doe.pem

# Run the non-SGX demo client
./demo/bin/ecdh-client -r demo/data/client_priv_test.pem -u demo/data/server_cert_test.pem -i localhost -p 7000

# Run the SGX demo client
./demo/bin/kmyth_sgx_retrieve_key_demo
```

sgx/TESTING.md
```diff
@@ -96,3 +94,27 @@ then the server does the same.
...
+
+The ECDHE key exchange format is identical to that used in the ECDHE test key server,
+which is described above.
+
+#### Usage
+
+The proxy usage is similar to the test key server application.
+
+```
+./demo/bin/tls-proxy -r data/server_priv_test.pem -u data/client_cert_test.pem -p LOCAL_PORT -I REMOTE_HOST -P REMOTE_PORT -C REMOTE_CA_CERT -R CLIENT_KEY -U CLIENT_CERT
```
PircDef, line 119: ```data/server_priv_test.pem``` and ```data/client_cert_test.pem``` should probably be SERVER_KEY and CLIENT_CERT  

Additionally, it's not clear to me what REMOTE, SERVER, and CLIENT would be in this context. Does ```-R CLIENT_KEY``` mean that the proxy server needs every client's private key?

sgx/demo/app/kmyth_sgx_retrieve_key_demo.c
```diff
@@ -51,7 +51,8 @@
 /* These parameters are hard-coded for now. */
 #define SERVER_IP "localhost"
 #define SERVER_PORT 7000
-#define KEY_ID "fake_key_id"
```
PircDef, line 54: ```fake_key_id``` was a good, descriptive string for the testing. Is there a compelling reason for moving to "7"?

sgx/TESTING.md
arp102: Here's one way to make those names more clear and consistent.

```
./demo/bin/tls-proxy -r ECDH_LOCAL_KEY -u ECDH_REMOTE_CERT -p ECDH_LOCAL_PORT -I TLS_REMOTE_HOST -P TLS_REMOTE_PORT -C TLS_REMOTE_CA_CERT [ -R TLS_LOCAL_KEY -U TLS_LOCAL_CERT ]
```

The `-R` argument is an RSA private key in PEM format, and the `-U` argument is its matching RSA public key PEM cert.
Both are required for authentication with PyKMIP, but they may not be required for other TLS services.

The `-U` cert must be signed by a CA, and the PyKMIP server must have a copy of the CA cert (the `ca_path` field in the PyKMIP server config).
It should be possible to use the same local key for both ECDH and TLS authentication, but I think it's better to keep them separate.

sgx/demo/app/kmyth_sgx_retrieve_key_demo.c
arp102: In PyKMIP, the unique key identifiers appear to be automatically-assigned database IDs, so they're decimal strings starting at `1` with the first key registered.

My PyKMIP setup instructions register 10 keys, so any value 1-10 would work for this test.

We could change the demo from get-by-id to get-by-name, but it would be a bit more complicated and probably require two separate KMIP requests.

arp102:
Updated to addressed some review feedback.

I forgot to include commit [055bbfb](https://github.com/NationalSecurityAgency/kmyth/pull/161/commits/055bbfbcb5ecbbc273691ca2402f8c7d2bcbce65) in the original PR,
which might have affected testing for the SGX demo with PyKMIP.

PR #164, opened 2022-04-20
Fix enclave_retrieve_key repeatability bug.

arp102:
The local private key parameter was being cleared, so it couldn't be re-used in other key requests.


JWZeta:
The code will allow for other key requests but it still has an issue with clean up of child processes.  The child processes are not being exited so they show up on the process list.

PR #144, opened 2021-11-04
Inclusion of PolicyOR

APysnack:
_note: resubmitted pull request after modifying the branch name._

This pull request implements an optional policyOR functionality that allows kmyth to seal/unseal against 2 different values of PCR's. Please feel free to contact with questions.

A formal demonstration of the functionality can be viewed here:
https://www.youtube.com/watch?v=ZCJS-8w6P90

If it assists the reviewer of the pull request, there are also about 2.5 hours of videos (all of which have "Kmyth" in the title) that discuss the majority of the steps that were taken in trying to solve this. Please note that we're all new to this technology, and the videos were partly meant to be informative, so they might be a bit lengthy at times (I recommend watching with 2x speed if they are).

If time permits in the future, there are a few things that could be refactored, and the policyOR could potentially be extended to more than just 2 policies (a max of 8).

Please let us know of any changes that may need to be made.

PR #166, opened 2022-06-02
Feature/add policy or for authorization

bit-belcher:
Duplicated PR#144 with edits and comments.


bit-belcher (commented):
Seems these changes are in order to expand to use of dual policyOR other than the convert_string_to_digest which was moved to utils/src/formattting_tools suggested in a later comment (https://github.com/NationalSecurityAgency/kmyth/pull/144#discussion_r755219449).

bit-belcher (commented):
also easy

bit-belcher:
Potential Changes for Next version(s) -- after completion of PR144/PR166 review.s
1. getkey function use: Could change to character-by-character approach according to a comment of bit-bender.
2. bool_policy_or is currently a uint_t and bool would be a better choice of type, so the name could change to policy_or. When the code is changed to handle this more sophisitcatedly this might be a good time to change to policy_or assuming it doesn't conflict with previous usages.
3. Names such as p_branch_1_64_data might also be changed to remove the numerical aspect but it looks like the 64 refers to the data type (base64 encoding) rather than the data length. The constant involves fixed 64-bit word so similar improvements might be warranted, or use of sizeof() constructions.
4. Add automated test of policies.dat infile/outfile capability (the infile is optional, check outfile production, do KAT).

bit-belcher:
List addressed by commits of PR144/PR166.

a. Much of the commits contain light refactoring, including expansion on comments. Some changes were even as minor as changing the name "expectedPolicy" to "expected_policy."

b. Code relocation as in assign_session_nonces function was relocated into kmyth_seal_unseal_impl.c, needed change from static type due to runtime considerations. Some logging functionality incorporated into rollNonces. Function convert_string_to_digest function moved to util/src/formatting_tools, addressing discussion_r755219449 from PR144.

c. Many changes involve division of dualizing policyOR, or elimination of unused cases. Many situations required adding an extra set of fields, for instance, and propagating variables in function calls down their call stacks. For instance, change c459a9... changed a name in a function call, so we tracked down all usages of this in the call stack to check consistency.

d. Unseal capability augmented, including a two-argument infile/outfile.

e. Type use changed to use of TMP2B_DIGEST (TPM2B_AUTH was previous). assign_session_nonces functionality also replaced by create_auth_session. Another point has zeroization of pHashList which is done in a loop and which could be done via memcpy with sizeof(TPM2B_DIGEST) method which might be more desirable.

f. 'Magic numbers' have generally been removed or converted to use of constants.

src/tpm/kmyth_seal_unseal_impl.c
```diff
@@ -151,10 +154,82 @@ int tpm2_kmyth_seal(uint8_t * input,
     return 1;
   }
 
+  // optional argument allowing user to receive a hex dump of the policy digest,
+  // to be used to calculate a second policy for policyOR authorization
+  if (bool_trial_only == 1)
```
bit-bender, line 159: I looks like kmyth-reseal (and kmyth_seal -g) forces this conditional to be executed.

While on a first pass this prints out a hex string to be used as an input parameter for the 'reseal', I don't understand how it is disabled on the subsequent call to kmyth-reseal where the printed out hex string is passed as an input parameter. In that case, I would not think we want the early termination, but instead produce a new .ski file.

Also, I think that the hex string printed out reflects the entire TPM2_DIGEST struct. The first four hex characters are '0200', which I think reflects the little endian representation of the digest size (0x0020 = 32). I think that we should specify only the hex string for the hash itself as a command line input parameter.

utils/src/formatting_tools.c
```diff
@@ -316,3 +316,78 @@ int concat(uint8_t ** dest, size_t * dest_length, uint8_t * input,
...
+{
+  // total number of hex values in the TPM2B digest
+  size_t digest_size = (8 * sizeof(digest)) + 2;
+
+  char * hex_buf;
+
+  hex_buf = (char *) malloc( digest_size+1 );
+  if( hex_buf == NULL ) return 1;
+
+  memcpy(hex_buf, digest, digest_size);
```
bit-bender, line 371: Since the variable 'digest' points to a TPM2B_DIGEST struct, I think that we should be using the formatting provided by the struct to parse out only the SHA-256 digest value.

Again, I do not think that we need to create new variables 'hex_buf' and 'digest_size', when the values of the hash size and hash result are already directly available as digest->size and digest->buffer respectively.

In the for loop that produces the hex string, I think we could iterate over the hash value as:
```
for (int i = 0; i < digest->size; i++)
{
...
}
```
and access hash bytes as

```
digest->buffer[i]
```

This would also change the behavior of the code as the size bytes would no longer be included in the hex string, because we would not be computing it over all of the bytes in the struct. In my opinion, not including the length bytes is clearer and negates the need to consider endian-ness.

If we do want a new buffer, I think the memcpy should be revised to something like

```
memcpy(hex_buf, digest->buffer, digest->size);
```

bit-bender (changes requested):
I'm still reviewing and experimenting with these changes, but I wanted to release some initial comments.

include/defines.h
```diff
@@ -308,4 +308,12 @@
  */
 #define KMYTH_GETKEY_RX_BUFFER_SIZE 256
 
+/**
+ * 
+ * Length of hexadecimal string of the policy digest read within TPM 2.0.
+ * (in characters).
+ *
+ */
+#define POLICY_DIGEST_HEXSTRING_SIZE 132
```
bit-bender, line 317: I think this number is based on incorrect logic. I think that the hex size for a policy digest should be twice the size of the number of hash bytes produced by the algorithm used. For SHA-256, for example, which produces a 32-byte hash result, I think the hex string should be 64 bytes (plus a string terminator byte, perhaps).

src/tpm/kmyth_seal_unseal_impl.c
```diff
@@ -151,10 +154,82 @@ int tpm2_kmyth_seal(uint8_t * input,
...
+  {
+    // size of string is 128 chars for the size of the TPM2B_DIGEST buffer + 4
```
bit-bender, line 161: I think this is wrong. The size of the TPM2B_DIGEST struct is two bytes for its 'size' member and for the current kmyth parameters, it is 32 bytes for the 'buffer' member of that struct. 

I believe that we should print out only the contents of the TPM2B_DIGEST struct, which should require 64 hex characters (or, more portably, twice the value contained in the 'size' member).

Currently, this prints out 4 hex characters for the size concatenated with 64 hex characters for the digest (hash result) padded with 64 zero hex characters.

bit-belcher: Addressed in subsequent commits.

src/tpm/kmyth_seal_unseal_impl.c
bit-bender: The comment about the kmyth-reseal workflow was addressed by creating an issue (#171) so that this can be addressed by a future revision.

The comment about the format in which the hex string is printed out was addressed by changes in commit [5ae6a9a](https://github.com/NationalSecurityAgency/kmyth/pull/166/commits/5ae6a9a7e73aa09ed68443fb5be63a6c32741480).

utils/src/formatting_tools.c
```diff
@@ -316,3 +319,95 @@ int concat(uint8_t ** dest, size_t * dest_length, uint8_t * input,
...
+    return 1;
+  }
+
+  if (digest == NULL || digest->buffer == NULL )
+  {
+    kmyth_log(LOG_ERR, "invalid digest argument ... exiting");
+    return 1;
+  }
+
+  // initializes buffer with all 0 hex values
```
bit-bender, line 346: I think the code here is fine, but the comment no longer matches (i.e., expectedPolicyBuffer is allocated, but not initialized to all-zero hex values)

PR #169, opened 2022-09-29
Update code based on new SGX compile flags

JWZeta:
These changes correspond with the pelz code changes based on the SGX compile flags added to the pelz Makefile.

### DIFF
--- a/sgx/untrusted/include/ocall/ecdh_ocall.h
+++ b/sgx/untrusted/include/ocall/ecdh_ocall.h
@@ -40,6 +40,8 @@ extern "C"
  * @param[in]  server_host                String IP address or hostname used to
  *                                        connect to the key server.
  *
+ * @param[in]  server_host_len            Length (in bytes) of server_host string.
+ *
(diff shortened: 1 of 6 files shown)


sgx/untrusted/include/ocall/ecdh_ocall.h
```diff
@@ -51,7 +49,7 @@ extern "C"
  *
  * @return 0 on success, 1 on failure
  */
-  int setup_socket_ocall(const char *server_host, int server_host_len,
+  int setup_socket_ocall(const char *server_host,
```
bit-bender, line 52: This change to the setup_socket_ocall() API breaks the demo code, which uses it. You have to also change the specification of this OCALL in the .edl file. The problem there, is that we specify the size of the 'server_host' string as 'count = server_host_len' and the SDK uses this for memory allocation. Since it is 'const char *', I tried not specifying the size, but that didn't work. I also believe that you cannot use something like strlen() in the .edl. What is the error that passing the length parameter causes?

bit-bender (approved):

PR #176, opened 2022-12-15
Corrected 'retrieve key' demo KMIP request/response exchange between proxy and demo server

bit-bender:
In the 'retrieve key' demo proxy and demo-server code, prior to sending KMIP requests (from the proxy to the demo-server) and KMIP responses (returned from the demo-server to the proxy). a two-byte length value was sent to facilitate allocating a properly sized receive buffer. This approach, however, is not consistent with a true KMIP server, however. This pull request modifies the code to exchange the KMIP requests and responses only and unmodified, in an attempt to be compliant with an actual KMIP server.

This pull request renames the `get_key_from_tls_server()` function to a more generic `get_resp_from_tls_server()` and re-uses that code in the proxy's `proxy_get_kmip_response()` function. The KMYTH_GETKEY_RX_BUFFER_SIZE  constant was increased to 16384 and a check to error if this buffer is ever completely filled was added.

### DIFF
--- a/src/network/tls_util.c
+++ b/src/network/tls_util.c
@@ -434,7 +434,7 @@ int get_resp_from_tls_server(BIO * bio,
   {
     kmyth_log(LOG_ERR, "receive buffer full (%d bytes) ... exiting", recv);
     free(buf);
-    return 1;t s
+    return 1;
   }
 
   *resp_size = recv;
--- a/src/network/tls_util.c
+++ b/src/network/tls_util.c
@@ -388,9 +388,9 @@ int tls_set_context(unsigned char *client_private_key,
 //############################################################################
 // get_key_from_tls_server()
 //############################################################################
-int get_key_from_tls_server(BIO * bio,
-                            char *message, size_t message_length,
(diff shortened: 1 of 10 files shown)


src/network/tls_util.c
```diff
@@ -430,17 +430,23 @@ int get_key_from_tls_server(BIO * bio,
     return 1;
   }
 
-  *key_size = recv;
+  if (recv >= buf_size)
+  {
+    kmyth_log(LOG_ERR, "receive buffer full (%d bytes) ... exiting", recv);
+    free(buf);
+    return 1;t s
```
JWZeta, line 437:
```suggestion
    return 1;
```

JWZeta:
Getting error ```/usr/local/bin/ld: demo/obj/tls_proxy.o: in function `proxy_handle_session': 
tls_proxy.c:(.text+0x27c): undefined reference to `get_resp_from_tls_server'```.
Need to add ```network/tls_util.c``` to Makefile.

JWZeta (approved):

Issue #172, opened 2022-10-11
kmyth-seal with -g (get expected policy digest) option should not produce empty .ski file


bit-bender:
As part of the Feature/add policy or for authorization pull request (PR #166), an option to print out (get) the policy digest was added. The application terminates normally after printing this result but also produces an empty .ski output file. In this case, it would be cleaner to perform this new functionality without producing the empty output file (.ski) result.

bit-bender:
Resolved by PR #181.

PR #181, opened 2023-02-18
Fixed empty file creation when -g option is passed to kmyth-seal (#172)

Costinteo:
The change is pretty simple and self-explanatory. Issue #172 seems to happen because of `write_bytes_to_file` being called regardless of `bool_trial_only`. So I've added a simple guard before it.

Tested it and it seems like the code addition indeed fixes the empty file creation.

### DIFF
--- a/src/main/seal.c
+++ b/src/main/seal.c
@@ -388,13 +388,17 @@ int main(int argc, char **argv)
   kmyth_clear(authString, auth_string_len);
   kmyth_clear(ownerAuthPasswd, oa_passwd_len);
 
-  if (write_bytes_to_file(outPath, output, output_length))
+  // only create output file if -g option is NOT passed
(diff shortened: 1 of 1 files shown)


bit-bender (approved):
Thank you for contributing this simple and elegant fix.

This addresses what I believe to be the biggest issue with use of this option (i.e., creation of extraneous zero-sized files). In my opinion, though, the -g (or --get_exp_policy) option is still handled a little awkwardly. For example, it still requires an input file, even though for this functionality it is not technically needed, it still requires an output filename (or creates/uses a default filename that doesn't collide with an existing file), the policy digest hex string has no label and is missing a newline at the end, ...

I mention this only to let you know that, at some point, we may create a follow-on issue to address the handling of this option in a broader fashion.

Thanks again.

Issue #173, opened 2022-10-11
Detect TPM "Policy-OR" criteria from .ski file (remove "policy OR" command line option for kmyth-unseal)


bit-bender:
Feature/add policy or for authorization (PR #166) added the capability to kmyth-seal data that can be unsealed if the policy digest matches one of two possible criteria (policy-OR). When unsealing, however, the user must currently specify the -p (policy OR) flag. As the existence of a policy OR criteria should be easily detectable when parsing/reading the input .ski file, however, the user should not have to specify this as a command line option, but instead kmyth-unseal should just apply whatever authorization criteria is specified in the .ski file.

PR #183, opened 2023-03-21
Updated kmyth-reseal

fryerT:
Reseal now unseals a .ski file, seals the unsealed file, and then deletes the unsealed file. '-o' output now defaults to a .txt file

### DIFF
--- a/src/main/reseal.c
+++ b/src/main/reseal.c
@@ -336,19 +336,18 @@ int main(int argc, char **argv)
     }
   }
 
-  uint8_t *unseal_output = NULL;
-  size_t unseal_output_len = 0;
-
   int *pcrs = NULL;
   int pcrs_len = 0;
 
   if (parse_pcrs_string(pcrsString, &pcrs, &pcrs_len) != 0 || pcrs_len < 0)
   {
     kmyth_log(LOG_ERR, "failed to parse PCR string %s ... exiting", pcrsString);
-    free(unseal_output);
     free(pcrs);
     return 1;
   }
+
+  uint8_t *unseal_output = NULL;
+  size_t unseal_output_len = 0;
 
  // Call top-level "kmyth-unseal" function
   if (tpm2_kmyth_unseal_file(inPath, &unseal_output, &unseal_output_len,
@@ -356,10 +355,11 @@ int main(int argc, char **argv)
                              (uint8_t *) ownerAuthPasswd, oa_passwd_len,
                              bool_policy_or))
   {
+    kmyth_log(LOG_ERR, "kmyth-unseal error ... exiting");
     kmyth_clear_and_free(unseal_output, unseal_output_len);
-    kmyth_log(LOG_ERR, "kmyth-unseal failed ... exiting");
     kmyth_clear(authString, auth_string_len);
     kmyth_clear(ownerAuthPasswd, oa_passwd_len);
+    free(pcrs);
     return 1;
   }
   
@@ -387,6 +387,7 @@ if (tpm2_kmyth_seal(unseal_output, unseal_output_len, &seal_output, &seal_output
   }
 
   kmyth_clear_and_free(unseal_output, unseal_output_len);
+  free(pcrs);
   kmyth_clear(authString, auth_string_len);
   kmyth_clear(ownerAuthPasswd, oa_passwd_len);
 
@@ -396,6 +397,8 @@ if (tpm2_kmyth_seal(unseal_output, unseal_output_len, &seal_output, &seal_output
   strncat(renamePath, ".orig", 5);
   if (!stat(renamePath, &st) && !forceOverwrite)
   {
+    free(seal_output);
+    free(renamePath);
     kmyth_log(LOG_ERR,
               "output filename (%s) already exists ... exiting",
               renamePath);
--- a/src/main/reseal.c
+++ b/src/main/reseal.c
@@ -133,8 +133,8 @@ static void usage(const char *prog)
           " -i or --input           Path to file containing the data to be sealed.\n"
           " -o or --output          Destination path for the sealed file. Defaults to <filename>.ski in the CWD.\n"
           " -f or --force           Force the overwrite of an existing .ski file when using default output.\n"
-          " -p or --pcrs_list       List of TPM platform configuration registers (PCRs) to apply to authorization policy.\n"
-          "                         Defaults to no PCRs specified. Encapsulate in quotes (e.g. \"0, 1, 2\").\n"
+//          " -p or --pcrs_list       List of TPM platform configuration registers (PCRs) to apply to authorization policy.\n"
+//          "                         Defaults to no PCRs specified. Encapsulate in quotes (e.g. \"0, 1, 2\").\n"
           " -c or --cipher          Specifies the cipher type to use. Defaults to \'%s\'\n"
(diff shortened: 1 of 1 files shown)


src/main/reseal.c
```diff
@@ -397,8 +474,21 @@ int main(int argc, char **argv)
     return 1;
```
bit-bender, line 474: I think the previous two lines (472-473) may no longer be needed (assuming other suggestions are implemented)

Remove `free(output)` - replaced by `free(seal_output)` in the previous suggestion.

Remove `free(pcrs)` - freed earlier in a suggestion above.

bit-bender (changes requested):
Please review and consider these comments/suggestions.

src/main/reseal.c
```diff
@@ -407,31 +404,38 @@ int main(int argc, char **argv)
...
   }
-  // We are done with authString and ownerAuthPasswd, so clear them
-  //kmyth_clear(authString, auth_string_len);
-  //kmyth_clear(ownerAuthPasswd, oa_passwd_len);
 
-  if (stdout_flag == true)
+  // rename the input file **** This is temporary. Needs to be changed slightly
+  int worked;
+  char *newFileName = "orginalFile.ski";
```
bit-bender: This renames the input to a "fixed" file name. If we run `kmyth-reseal` on multiple .ski input files, we will have to rename this result between every invocation of `kmyth-reseal` or the names will conflict. Note: this is also missing the code to check for that filename collision.

Is there a reason that you think this is better than the proposed solution that renamed the input file to <input filename>.orig?

src/main/reseal.c
```diff
@@ -135,6 +135,7 @@ static void usage(const char *prog)
           " -f or --force           Force the overwrite of an existing .ski file when using default output.\n"
           " -p or --pcrs_list       List of TPM platform configuration registers (PCRs) to apply to authorization policy.\n"
           "                         Defaults to no PCRs specified. Encapsulate in quotes (e.g. \"0, 1, 2\").\n"
+          " -g                      Used when resealing a file that already contains a policy OR\n" // may change later
```
bit-bender, line 138: So, my original thought was to initially assume (i.e. require) that the input file to be 'resealed' would not have a policy-OR authorization criteria.

The real fix to this (i.e., the need for a command-line option like this) is the update that addresses Issue #173. When code to address that issue is merged, the detection of a policy-OR criteria from the .ski file format will be automatic and no longer require a command-line option.

I think inclusion of this option, for now, although kind of clunky, makes `kmyth-reseal` more generic, which is a good thing.

src/main/reseal.c
```diff
@@ -299,207 +295,130 @@ int main(int argc, char **argv)
...
+                      bool_trial_only))
   {
     kmyth_log(LOG_ERR, "kmyth-seal error ... exiting");
     kmyth_clear(authString, auth_string_len);
     kmyth_clear(ownerAuthPasswd, oa_passwd_len);
+    kmyth_clear_and_free(unseal_output, unseal_output_len);
     free(pcrs);
-    free(outPath);
-    free(output);
+    free(unseal_output);
```
bit-bender, line 386: On line 384, this is freed by the kmyth_clear_and_free() call

src/main/reseal.c
```diff
@@ -299,207 +295,130 @@ int main(int argc, char **argv)
...
-*/
-
-
-
-  free(pcrs);
-  free(outPath);
-  free(output);
-  //free(tempOut);
 
+  kmyth_clear_and_free(unseal_output, unseal_output_len);
```
bit-bender, line 420: This is very minor, but, since we no longer need `unseal_output` after the the tpm2_kmyth_seal() call, I think we could move this earlier.

src/main/reseal.c
```diff
@@ -242,8 +246,9 @@ int main(int argc, char **argv)
     case 'e':
```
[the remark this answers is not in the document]
bit-bender: The -g option was used for a different purpose.

src/main/reseal.c
```diff
@@ -279,126 +282,141 @@ int main(int argc, char **argv)
...
-                           bool_trial_only))
+ // Call top-level "kmyth-unseal" function
+  if (tpm2_kmyth_unseal_file(inPath, &unseal_output, &unseal_output_len,
+                             (uint8_t *) authString, auth_string_len,
+                             (uint8_t *) ownerAuthPasswd, oa_passwd_len,
+                             bool_policy_or))
   {
-    kmyth_log(LOG_ERR, "kmyth-reseal error ... exiting");
+    kmyth_clear_and_free(unseal_output, unseal_output_len);
+    kmyth_log(LOG_ERR, "kmyth-unseal failed ... exiting");
```
bit-bender, line 360: This is an extremely trivial comment, but, everywhere else, we log the error and then clean-up/return. For consistency only, I would move the `kmyth_log()` statement up one line to the beginning of the code block for this conditional.

src/main/reseal.c
```diff
@@ -279,126 +282,141 @@ int main(int argc, char **argv)
...
     kmyth_clear(authString, auth_string_len);
     kmyth_clear(ownerAuthPasswd, oa_passwd_len);
+    return 1;
```
bit-bender, line 363: I think this block needs to also `free(pcrs)`

src/main/reseal.c
```diff
@@ -279,126 +282,141 @@ int main(int argc, char **argv)
...
+  strncat(renamePath, ".orig", 5);
+  if (!stat(renamePath, &st) && !forceOverwrite)
   {
-    kmyth_log(LOG_ERR, "error writing data to .ski file ... exiting");
-    free(outPath);
-    free(output);
-    free(pcrs);
+    kmyth_log(LOG_ERR,
+              "output filename (%s) already exists ... exiting",
+              renamePath);
```
bit-bender, line 401: I think we need to `free(seal_output)` and `free(renamePath)` before returning due to this error.

src/main/reseal.c
```diff
@@ -279,126 +282,141 @@ int main(int argc, char **argv)
...
+    kmyth_clear_and_free(unseal_output, unseal_output_len);
     free(pcrs);
-    free(outPath);
-    free(output);
     return 1;
   }
 
+  kmyth_clear_and_free(unseal_output, unseal_output_len);
   kmyth_clear(authString, auth_string_len);
   kmyth_clear(ownerAuthPasswd, oa_passwd_len);
```
bit-bender, line 391: I think we neglect to `free(pcrs)` in the normal termination case. I think we are done with it from here on, so could add it here.

bit-bender (changes requested):
A few additional comments after reviewing the latest code again.

src/main/reseal.c
```diff
@@ -279,126 +273,144 @@ int main(int argc, char **argv)
...
+  strncat(renamePath, ".orig", 5);
+  if (!stat(renamePath, &st) && !forceOverwrite)
   {
-    kmyth_log(LOG_ERR, "error writing data to .ski file ... exiting");
-    free(outPath);
-    free(output);
-    free(pcrs);
+    free(seal_output);
+    free(renamePath);
+    kmyth_log(LOG_ERR,
```
bit-bender, line 393: This kmyth_log() statement should be moved to the beginning of this block. I prints garbage for the output filename because we free `renamePath` and then try to use it.

PR #185, opened 2023-11-14
Bugfix/string handling warnings

bit-bender:
This pull request addresses compiler warnings that highlighted incorrect string handling:

- In a couple cases, the strncat() function was passed a bound that was the length of the source string being appended to the destination string. As this is essentially the behavior of strcat(), the strncat() calls could have been changed to strcat() calls to address the compiler warnings. In this pull request, however, the choice was made to instead increase the bound to to make the copied source string null terminated.

- The maximum default filename length was only 25 characters. This pull request increases the maximum length to 64.
- An incorrect string comparison was fixed.
- An extra byte was allocated for `outPath_size` to include space for null termination

### DIFF
--- a/src/main/seal.c
+++ b/src/main/seal.c
@@ -303,20 +303,15 @@ int main(int argc, char **argv)
   if (outPath == NULL && !boolTrialOnly)
   {
     // create buffer to hold default filename derived from input filename
-    //    size = string length of basename +
-    //           string length of delimiter ('.') +
-    //           string length of file extension +
-    //           extra byte for null terminator
-    char default_fn[KMYTH_MAX_DEFAULT_FILENAME_LEN +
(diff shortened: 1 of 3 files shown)


src/main/seal.c
```diff
@@ -269,7 +269,13 @@ int main(int argc, char **argv)
   // Some options don't do anything with -g, so warn about that now.
   if(boolTrialOnly)
   {
-    if(authString != NULL || ownerAuthPasswd != "" || outPath != NULL || inPath != NULL || cipherString != NULL || forceOverwrite || expected_policy != NULL)
+    if(authString != NULL ||
+       (strcmp(ownerAuthPasswd, "") != 0) ||
```
b-carr, line 273: I think it'd be cleaner to use ```strnlen(ownerAuthPassword, 1) != 0``` here.

bit-bender: Yes, done.

PR #187, opened 2023-11-16
Refactor PCR-based policy-OR functionality

bit-bender:
This pull request is an attempt to improve the `kmyth-seal` and `kmyth-reseal` functionality addressing PCR-based policy-OR criteria:

- The existing approach uses a different .ski format when sealing with a PCR-based policy-OR criteria. This pull request standardizes on a 'PCR SELECTIONS' block encoding a `PCR_SELECTIONS` struct containing a list of 'TPML_PCR_SELECTION` structs and a `POLICY OR` block encoding a `TPML_DIGEST` struct containing a list of policy digests for all scenarios and negates the need for a command-line parameter to specify the format (addresses kmyth issue #173).
- This pull request enables automatically including the PCR selection and policy digest data recovered from the input .ski file in a `kmyth-reseal` result. The approach taken, however, required significant changes to the kmyth API.
- This pull requests attempts to allow the user to specify a PCR-based policy-OR criteria in a clearer and more meaningful way. In the approach taken, the user now expresses an "expected policy" criteria as a string specifying a pair of PCR selection and policy digest values.
- For the existing approach, only two policy digests can be specified, and both must use the same PCR selections. This pull request makes this more flexible, allowing up to eight policy-OR criteria, each with independent PCR selections.
- Changes to the API required changes to a number of tests. Some testing for new or revised functionality was added, but making the testing more complete is probably still a future goal.
- Policy-OR criteria in kmyth are currently only PCR based. In the future, functionality could potentially be extended to allow more flexible and generic policies. This code, though, enforces that all policies in a policy-OR criteria are PCR-based. For the current kmyth, there is no value in using a policy-OR criteria if any of the criteria is not PCR-based.


src/tpm/kmyth_seal_unseal_impl.c
```diff
@@ -143,93 +140,96 @@ int tpm2_kmyth_seal(uint8_t * input,
...
   if (bool_trial_only == true)
   {
-    // size of string is 2x chars for the size of the TPM2B_DIGEST buffer + 4
-    // for TPM2B struct which encodes size in memory + 1 byte for null termination.
-    // prints to the console and finishes the program without sealing
-    // By using TPM2B_DIGEST size we're reserving a little bit more than needed but
-    // then the output_string can be reserved based on compile-time values.
-    size_t string_size = (2 * sizeof(TPM2B_DIGEST)) + 1;
-    char output_string[string_size];
+    // create stack variable for digest hex-string
```
b-carr, line 173: Why does it matter that it's on the stack? I have no objection, just not sure why that's called out in the comment.

src/tpm/kmyth_seal_unseal_impl.c
```diff
@@ -523,36 +621,35 @@ int tpm2_kmyth_unseal(uint8_t * input,
...
     }
 
     if (read_bytes_from_file(input_path, &data, &data_len))
     {
-      kmyth_log(LOG_ERR, "seal input data file read error ... exiting");
+      kmyth_log(LOG_ERR, "seal input data file read error");
       if (data != NULL)
       {
-	free(data);
+	      free(data);
```
b-carr, line 652: Extra whitespace.

test/src/tpm/tpm2_interface_test.c
```diff
@@ -692,29 +691,192 @@ void test_apply_policy(void)
...
+    {
+      // create policy digest for second policy branch
+      CU_ASSERT(create_policy_digest(sapi_ctx,
+                                     &(pcrs_struct_3.pcrs[1]),
+                                     &pOR_digests_struct,
+                                     &policy2) == 0);
+      CU_ASSERT(policy2.size == KMYTH_DIGEST_SIZE);
+      char policy2_str[(KMYTH_DIGEST_SIZE * 2) + 1];
+      convert_digest_to_string(&policy2, policy2_str);
+      CU_ASSERT(strncmp(policy1_str,
```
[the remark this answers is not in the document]
bit-bender: I believe that the issue here is a version dependency on TPM2 Tools. I have rewritten the tests for apply_policy() and apply_policy_or() to not require these system calls and, hence, eliminate the dependency on TPM2 Tools.

Also, in looking at these tests, I realized that they were written for the current policy-OR implementation and possibly only slightly modified because of the changes in this pull request. Therefore, I completely refactored the tests for apply_policy() and apply_policy_or() in an attempt to better exercise, test, and document this functionality.

src/tpm/kmyth_seal_unseal_impl.c
bit-bender: Yes, it really does not matter. Since it is a temporary buffer that exists only to facilitate printing out the hex-string representing the policy digest, I decided that a stack variable, that didn't require explicit memory allocation and freeing would be a good choice here, but, it is certainly not a required or important that this variable is implemented that way. I have reworded the comment to attempt to emphasize instead that it is a temporary variable (i.e., only exists within the scope of the if statement).

src/tpm/kmyth_seal_unseal_impl.c
bit-bender: I think I fixed this. I looked fine in my editor (VS Code), both before and after commit [8bc5da2](https://github.com/NationalSecurityAgency/kmyth/pull/187/commits/8bc5da26598a0cc0959dc0222aa71e4d48d29c4d), but 'git diff' indicated the removal of the extra whitespace.

bit-bender:
While attempting to address the comments, I stumbled across some other issues that I attempted to correct:
- Both an init_policyOR() and an init_policy_or() function were present in the code. As init_policy_or() was unused, I cleaned up this dead code.
- The `kmyth-reseal` main() used init_policyOR() to configure the PCR selections list and policy digest list structs for a policy-OR authorization criteria. The `kmyth-seal` main(), however, did not. I attempted to update both to use the policy_OR() function in a similar fashion.
- As there were no tests for parse_exp_policy_string_pairs() and init_policyOR(), I added tests for this functionality.
- I discovered that the parse_exp_policy_string_pairs() function used "," as the string pair delimiter in one of the `strtok()` calls and ";" in another. The "," case is invalid because it interferes with its use as a PCR list delimiter. In testing, however, I decided that the choice of ";" as a string pair delimiter is also not ideal. It requires encapsulating the string in quotes to avoid it being decoded as a command delimiter on the bash command line. Therefore, I changed the delimiter to be a forward slash ("/").
- I corrected an indexing error in the init_policyOR() code populating the policy digest list struct.
- I filled in some missing parameter descriptions for init_policyOR() in the tpm2_interface.h header file
- I expanded the parse_exp_policy_string_pairs() code to handle the presence of a '0x' prefix for the hexadecimal string representing a policy digest (a hex-string without the prefix also still works).

bit-bender:
Addressed a merge conflict in src/main/seal.c.

b-carr (approved):

PR #197, opened 2024-11-20
Add -fPIE option to SGX C++ file compilation / Address TLS issue with SGX ECDH demonstration

bit-bender:
The SGX ECDH demonstration code would not build under Red Hat Enterprise Linux (RHEL) 9. It generated build errors when processing .cpp sources and suggested adding the position independent code for executable (-fPIE) compile option. This pull request, therefore, addresses the build issue by adding the '-fPIE' option to the kmyth 'SGX_COMMON_CXX_FLAGS' Makefile variable. This must be necessary for RHEL 9 because of something like Address Space Layout Randomization (ASLR) features not present in my previous Ubuntu development environment. I did verify that this modification both enabled the build under RHEL 9 and did not break the build on an Ubuntu 22 platform.

After getting the demonstration code to build, the demonstration was failing due to TLS certificate verification errors (unknown CA). As I was on a machine with a newer version of OpenSSL, I looked into the test key/certificate generation and discovered that the X509 extensions were not being set when the test certificates were generated. I fixed this issue by refactoring the 'gen_test_keys_certs.bash' in the kmyth SGX ECDH demonstration data directory:

- Instead of assigning the X509 extensions within the configuration file, this pull request proposes switching to use the command-line '-extensions' option to specify the section in the configuration file that contains the specified extension settings.

- Instead of one configuration file per demo mode, this pull request proposed consolidating the configuration for all three demo nodes into a single file.

- This pull requests proposes modifying test key/certificate generation to a one-step approach. Rather than generating the private key separately, a Certificate Signing Request (CSR), the certificate itself, and a serial number file for each generated certificate, a single 'openssl req' command is used to generate matched private key and certificate files. Therefore, fewer test files are created and need to be cleaned up when no longer needed.

Finally, this pull request proposes changes intended to clean up some 'echo' statements in the Makefile to address minor formatting issues with the demo display output.


bit-bender:
Thank you for the confirmation that this builds on Fedora as well. That is a very helpful data point.

Also, thank you for the feedback about the incomplete installation instructions. The valgrind API dependency was added in PR #195 and noted in INSTALL.md, but not included in the installation examples. I have added a commit which adds the 'valgrind-devel' package to the Red Hat command-line package installation example. For Ubuntu, I believe that 'valgrind-dev' was replaced by the 'valgrind' package. I also verified that on my Ubuntu machine I currently only have the 'valgrind' package installed.

b-carr (approved):

